PR #14, opened 2014-06-11
Added methods to remove nodes from the topology

floxyz:
- The NetworkConfigurator now offers a way to delete nodes from the topology and add them.
- The ARP protocol does not crash anymore in case a node does not exist anymore


avarga:
commit "Added methods to remove nodes from the topology": it's not so simple...

PR #16, opened 2014-06-25
Include probabilistic obstacles attenuation

aarizaq:
Probabilistic obstacle attenuation


avarga:
This pull request covers 3 unrelated topics, could we split it to 3 different pull requests?

For the video trace model: is there some documentation what kind of video stream it models (codec etc), what is the input file format, and how to generate such file from a real video file?

rhornig:
@sommer The Traci patches in this pull request seem to be quite outdated. I guess we should not merge this. What do you think?

rhornig:
Color support is no longer needed as the new Canvas API in OMNeT++ 5.0 covers this.

rhornig:
The obstacle model was totally rewritten for INET 3.0 so the probabilistic obstacles attenuation patch seems to be obsolete.

rhornig:
The video tracing library might be interesting, but it needs proper documentation as in its current state it's very hard to figure out what it models (and how to use it). Perhaps it would be ok, we proper documentation on top of the current integration branch in s separate pull request.

rhornig:
I recommend closing this pull request without merging.

sommer:
The TraCI patches are seriously outdated. I suggest to skip them.

rhornig:
@aarizaq Is it ok to close this without merging? D you still want to merge the Video Stream library part? Do you have some extra info about it? (source/origin)

aarizaq:
The video stream can be interesting.

What documentation you need?

 

 

 

De: Rudolf Hornig [mailto:notifications@github.com] 
Enviado el: jueves, 19 de marzo de 2015 17:03
Para: inet-framework/inet
CC: Alfonso
Asunto: Re: [inet] Include probabilistic obstacles attenuation (#16)

 

@aarizaq <https://github.com/aarizaq>  Is it ok to close this without merging? D you still want to merge the Video Stream library part? Do you have some extra info about it? (source/origin)

rhornig:
@aarizaq Thanks, we have looked into the code. It seems that this is an extension of the video streaming client/server available in INET. As it duplicates its functionality the proper way to do it is to enhance the current implementation instead of adding a new module. The current video stream implementation is pretty bare bones, but we need to properly plan and document new features if we extend it. I was not able to figure out how these classes extend the current implementation what are the new parameters and how they affect the behavior.

rhornig:
The video tracing library might be interesting in a separate pull request, but only as a replacement/extension to the current video library + proper documentation of the behavior and parameters. The enhanced implementation should behave exactly like the old one using the default parameters.

aarizaq:
It works like the old module, I implemented it because I needed a realistic video pattern for an article (it was a condition of the reviewers), I thought to implement a synthetic video source but a colleague advised me to use the video trace library of the Arizona University, the advantage is that I can reproduce real video traffic.

A limitation of the implementation is that the video frames must be marked like type 'I', 'B' or 'P', other limitation is that it consume a lot of memory if the video trace file is big.

avarga:
So you ended up using the code and traces from http://trace.eas.asu.edu/? The code there is quite old (for OMNeT++ 3.x). If you ported it to a current INET, that could be of value too.

aarizaq:
What version do you need? I have the code for inet 2.6 and inet 2.99
I need to check if the code compile with the latest modification in inet

avarga:
For version 2.99 it would be great :) We'd check it in a topic branch and polish it a little, add docu, tests etc.

aarizaq:
I am going to check if the code compile with the latest inet modification and I will send a link with the code

avarga:
Great, thank you!

aarizaq:
the code for inet 2.99

https://github.com/aarizaq/inetmanet-2.0/blob/integration/src/inet/applications/udpapp
/VideoPacket.msg
https://github.com/aarizaq/inetmanet-2.0/blob/integration/src/inet/applications/udpapp/UDPVideoStreamSvr2.cc
https://github.com/aarizaq/inetmanet-2.0/blob/integration/src/inet/applications/udpapp/UDPVideoStreamSvr2.ned
https://github.com/aarizaq/inetmanet-2.0/blob/integration/src/inet/applications/udpapp/UDPVideoStreamCli2.h
https://github.com/aarizaq/inetmanet-2.0/blob/integration/src/inet/applications/udpapp/UDPVideoStreamCli2.cc
https://github.com/aarizaq/inetmanet-2.0/blob/integration/src/inet/applications/udpapp/UDPVideoStreamCli2.ned

avarga:
Good, thanks a lot!

PR #21, opened 2015-01-08
fix: remove radio from medium at destruction

riebl:
Without removal a SEGFAULT is caused in RadioMedium::computeMaxSpeed for dynamic radios, e.g. those controlled via TraCIScenarioManager


levy:
Thanks for the patch, it will be included in the next development release!

levy:
I should find another way to make this work for dynamic simulations, because the medium can't be used in the destructor. That's because the medium may have been already deleted when a non-dynamic simulation stops, but thanks, anyway!

PR #25, opened 2015-01-21
Serialize nr sack

ruengeler:
The cases SACK and NR_SACK have to be handled separately because struct nr_sack cannot be used to store both cases.


rhornig:
This pull request contains also unrelated code and changes from other pull requests. Could you please rewrite it to contain only the  c559698 commit? I could cheryy pick it, but I would love to take over all changes using a pull reuest, so other people will also clearly see what went into the integration barnch.

PR #27, opened 2015-01-21
Router alert handling (in integration branch)

pkerling:
This pull request includes an updated version of the topic/router-alert-handling branch modified for the integration branch.
See inet-framework/inet#10, inet-framework/inet#11

### DIFF
--- a/src/inet/applications/udpapp/UDPBasicApp.cc
+++ b/src/inet/applications/udpapp/UDPBasicApp.cc
@@ -86,6 +86,10 @@ void UDPBasicApp::setSocketOptions()
     if (receiveBroadcast)
         socket.setBroadcast(true);
 
+    bool routerAlert = par("routerAlert");
+    if (routerAlert)
+        socket.setRouterAlert(true);
+
     bool joinLocalMulticastGroups = par("joinLocalMulticastGroups");
     if (joinLocalMulticastGroups) {
         MulticastGroupList mgl = getModuleFromPar<IInterfaceTable>(par("interfaceTableModule"), this)->collectMulticastGroups();
--- a/src/inet/applications/udpapp/UDPBasicApp.ned
+++ b/src/inet/applications/udpapp/UDPBasicApp.ned
@@ -54,6 +54,7 @@ simple UDPBasicApp like IUDPApp
         int typeOfService = default(-1); // if not -1, set the ToS (IPv4) or Traffic Class (IPv6) field of sent packets to this value
         string multicastInterface = default("");  // if not empty, set the multicast output interface option on the socket (interface name expected)
         bool receiveBroadcast = default(false); // if true, makes the socket receive broadcast packets
+        bool routerAlert = default(false); // if true, sends packets with router alert option set
         bool joinLocalMulticastGroups = default(false); // if true, makes the socket receive packets from all multicast groups set on local interfaces
         @display("i=block/app");
         @signal[sentPk](type=cPacket);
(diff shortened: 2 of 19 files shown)


avarga:
Thank you for the patch! However, could you provide some details why this functionality is important enough to be included in INET? What is the motivation?

pkerling:
The motivation is that it is needed for some protocols (in this case, a layer-3 QoS path reservation protocol for a research project at a German university, but as far as I know also NSIS should it get implemented in INET) to be able to intercept IP packets from the network, possibly modify them and set a verdict (accept or drop packet). It could be argued that there should be a more general solution, a complete packet filtering framework. I agree in principle, but this would mean a massive amount of work and is probably not needed for most protocols. The router alert flag is a standardized IP flag that can be utilized for this purpose and using it seemed like a good compromise between too much effort and over-specialization to our needs.

pkerling:
See also https://www.rfc-editor.org/rfc/rfc2113.txt for a general motivation and https://www.rfc-editor.org/rfc/rfc5971.txt for its usage in NSIS

avarga:
When switching to NetFilter hooks, `IRouterAlertHandler.cc`/`h`, and `IPv4NetworkLayer.ned` modifications could also be dropped.

avarga:
> I think it could be implemented with NetFilter hooks for IPv4, but: The project is IPv6 only.
> IPv4 router alert handling was only added for the sake of completeness.

I see... How about we take the parts that deal with being able to send Router Alerts (ie. control info and IPv4/IPv6 datagram changes together with handling code), and you keep the handling of received Router Alerts (`IRouterAlertHandler`, etc) in you project-specific repo? See, there is no use case for router alert handling in INET right now (we don't have NSIS etc); when we implement NetFilter then it will also be suitable for hooking up router alert handling code, so the current `IRouterAlertHandler` stuff would only be temporary. And for unused temporary code the patch is a little heavy-weight (extra gates, extra modules, etc).  But I have no problem with accepting the part of *sending* router alerts, since it's quite straightforward, nonintrusive, goes in the direction we want INET to go, etc.

pkerling:
Sure, we'll definitely go with NetFilter once it lands in IPv6. We'd even help implementing once you say the IPv6 code is stable enough for that to be feasible. Do you have a general outline when this will be the case? With the release of INET 3, maybe?
I think integrating the router alert sending code would be great since it is a quick, easy and standardized way to tag packets for interception. As far as our concerns go though, we would not need it once the NetFilter code is in place. We can distinguish packets that need special handling by other means (UDP port is enough for us).

avarga:
Regarding the IPv6 code, it's not really about stability, but rather the code is currently ill-structured for maintainability and future extension. On IPv4,

avarga:
As for actions:
1. We can merge the router alert *sending* code, once the suggested modifications are done (i.e. the comments added to IPv4ControlInfo.msg and IPv6ExtensionHeaders.msg). Shall I wait for an updated pull request?
2. We can try "hack" the NetFilter support into the current IPv6 code if really needed, although it won't be that pretty. We originally decided to postpone IPv6 NetFilters to after the refactoring because it would be in the way for the refactoring.

pkerling:
1. Sure, I'll make a new request. Doesn't hurt to have the RA sending code in there.
2. We can use our patched INET until the IPv6 code is ready for NetFilter, so the issue is not time-pressing. I don't think it's any use adding more functionality at this point if a major rewrite is in the works. If you could just drop me a line when you're finished with that, we'd be happy to integrate NetFilter.

PR #29, opened 2015-01-23
Handle close command, i.e stop reading when the command is sent

ruengeler:


ruengeler:
That is on purpose. The commands of the case SCTP_E_SHUTDOWN should be performed as well.

rhornig:
If you intentionally do this, you must insert a /* fall through */ comment where the break would be. The reason is that coverity complains about MISSING_BREAK unless you manually annotate the switch case.

PR #31, opened 2015-01-28
Handle SACK and NR_SACK separately

ruengeler:
SACK and NR_SACK have to be handled separately because otherwise you get wrong results for numDups.
Also: getBitLength()/8 has been substituted by getByteLength()


rhornig:
Looks good. Reviewed by @rhornig @ZoltanBojthe

PR #34, opened 2015-01-28
Implement a virtual TUN interface including application and test

ruengeler:
The TUN interface connects the linklayer directly with the application. Thus, in a loop, packets can be sent from the application down the stack or via the TUN interface up the stack.
The application implements an echoing of packets. They are sent down the stack, delivered back to the application via the TUN interface and echoed back.
A test example verifies that all packets are reflected.


src/inet/linklayer/tun/TunInterface.cc
```diff
@@ -0,0 +1,97 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program; if not, see <http://www.gnu.org/licenses/>.
+//
+
+#define WANT_WINSOCK2
```
avarga: Is this needed? Probably just accidentally left in.

src/inet/linklayer/tun/TunInterface.cc
```diff
@@ -0,0 +1,97 @@
...
+
+#include <stdio.h>
+#include <string.h>
+
+#include <platdep/sockets.h>
```
avarga: Probably not needed

avarga:
We are currently looking at the code, more suggestions coming,..

levy:
I don't quite understand the TunBasicApp example. Maybe it's confusing for me because the TunBasicApp serves two purposes: it handles the packets of the interface, and it also communicates through the tunnel as a normal application would do.

I think it would be easier to understand if there would be a TunLoopbackApp which would simply send back the packets that it received from the interface. The example could use the standard UDP client/server applications that communicates through the tun interface backed by the TunLoopbackApp.

src/inet/node/inet/NodeBase.ned
```diff
@@ -120,6 +122,10 @@ module NodeBase
             parameters:
                 @display("p=149,406,row,10;q=txQueue;i=block/ifcard");
         }
+        tun: <default("TunInterface")> like ITunNic if hasTun {
```
ZoltanBojthe: tun interface must be a vector, and need one tunApp for each tun interface.
See it here: https://github.com/inet-framework/inet/commit/d1d72062e6c0a6cf15b61ea29569aff8e2f6c896

tuexen:
levy: The intention of the TunBasicApp is to have a simple test application to test that the tun interface is working. Nothing more, nothing less. It is not intended to be an example for a generic application using a tun interface. The application we have in mind (a version of packetdrill for INET) and will contribute later as a separate pull request will use a tun interface to inject (SCTP/TCP/UDP) packets and check if the corresponding actions at the transport layer happen. Therefore, it will also have an interface to each of the transport protocols which are supported for testing. I hope this gives you a clearer picture what we are after and why the test app looks like it is. So what do you suggest: Have a completely different app? Or no app at all (which I think is bad for testing) or chose a better name for the app to make clear that it is only for testing purpose...

levy:
tuexen: Thank you for your response. I think I understand what you are doing and why you are doing it. I agree that having tests is always better than not having. :-) On the other hand, we generally don't want to put test apps into the applications folder. They should rather go under tests/module/lib for example. For me the TunBasicApp name suggests it's a reusable application with some generic behavior for simulations that want to use tun interfaces. I think this is not the case and this could be also confusing for other users...

In fact, there are really two apps inside TunBasicApp. There's one that sends UDP packets and counts received UDP packets, and there's the tun interface part that echos received packets back. I think the first app already exists, and it's called UDPBasicApp (I think all required functionality is covered). If this part is removed, then what remains is something that is easy to understand. It could be called a TunLoopbackApp. It's even somewhat general in the sense that it effectively emulates a loopback interface using tun.

tuexen:
> On 10 Feb 2015, at 14:40, Levente Mészáros <notifications@github.com> wrote:
> 
> tuexen: Thank you for your response. I think I understand what you are doing and why you are doing it. I agree that 
Hi Levente,

OK, great.
> having tests is always better than not having. :-) On the other hand, we generally don't want to put test apps into the applications folder. They should rather go under tests/module/lib for example. For me the TunBasicApp name 
OK. Makes sense. I'll talk to Irene to see if the testapp can go there.
> suggests it's a reusable application with some generic behavior for simulations that want to use tun interfaces. I think this is not the case and this could be also confusing for other users...
Sure. So using a better name is also preferred. Will talk to her tomorrow.
> 
> In fact, there are really two apps inside TunBasicApp. There's one that sends UDP packets and counts received UDP packets, and there's the tun interface part that echos received packets back. I think the first app already exists, and it's called UDPBasicApp (I think all required functionality is covered). If this part is removed, then what remains is something that is easy to understand. It could be called a TunLoopbackApp. It's even somewhat general in the sense that it effectively emulates a loopback interface using tun.
Well, possibly. I think you can not only receive a packet, swap addresses and port numbers and
reinject, but you can inject a packet even with a non-local address and it would be routed.
I'll talk to Irene tomorrow to figure out how to change the test app.

PS: I hope this also shows up on GitHub. Not sure, I'm doing this the first time...

Best regards
Michael

levy:
Thank you for submitting this pull request! I merged it into the integration branch, let me know if there's anything wrong with it.

PR #36, opened 2015-01-30
TCP: fix bug #803 "TCP SACK issue". Author: Thomas Reschka

avarga:
Please review this commit.

### DIFF
--- a/src/inet/transportlayer/tcp/TCPConnectionUtil.cc
+++ b/src/inet/transportlayer/tcp/TCPConnectionUtil.cc
@@ -576,7 +576,7 @@ void TCPConnection::sendFin()
 
 void TCPConnection::sendSegment(uint32 bytes)
 {
-    if (state->sack_enabled && state->afterRto) {
+    if (state->sack_enabled) {
         // check rexmitQ and try to forward snd_nxt before sending new data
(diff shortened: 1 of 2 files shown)


avarga:
Under review by @scimbe

scimbe:
In process...

rhornig:
@scimbe Any news on this? If its ok, I would love to merge it.

scimbe:
Unfortunately I had no environment around to test it well. 
It will need minimum two more weeks, because I will get new hardware and software 
and I have to setup simulation pool to get an overview about the issue. 
Perhaps it is an alternative to merge this and to open a ticket assigned to me in the bug tracking tool... But that is up to you... 
This patch provides an improvement...

scimbe:
An additional push request?

Rudolf Hornig <notifications@github.com> schrieb am Mo., 23. Feb. 2015
17:40:

rhornig:
@scimbe  No, need to do anything special except please review the code change (single line) and whether i *correctly* solves the problem described in: http://dev.omnetpp.org/bugs/view.php?id=803

If you need more time, it's not a problem. We do not have the domain knowledge to decide whether this solution is ok or not. If you approve it, I will merge immediately.

PR #41, opened 2015-02-06
TCP scaling window patch

scimbe:
http://dev.omnetpp.org/bugs/view.php?id=810


scimbe:
Please let me know, if the format stuff is OK now

rhornig:
@scimbe Probably it would be good to include these changes also in the 'integration' branch. Changes in master will not be automatically merged into integration.

PR #42, opened 2015-02-08
removed depricated register to resolve the warnings with LLVM version 6....

tillsteinbach:
...0 when compiling using c++11 (see also http://en.cppreference.com/w/cpp/language/storage_duration). This should be done in OMNeT++ as well!

Best regards
Till


avarga:
Merge manually. I edited the commit message to suit the following format:

<component-or-topic>: <very-brief-description>
<blank-line>
<details, possibly on several lines>


"INETMath.h: eliminated warning
    
removed depricated register to resolve the warnings with LLVM version 6.0 when compiling using c++11 (see also http://en.cppreference.com/w/cpp/language/storage_duration). This should be done in OMNeT++ as well"

PR #43, opened 2015-02-09
Follow up of pull request #42

tillsteinbach:
Finding "register" in a framework that deals with protocols is hard :)
I found two more occurrences of "register" that I missed in pullrequest #42


rhornig:
merged

PR #44, opened 2015-02-16
Flush cached reception intervals

riebl:
Cache with reception intervals never got flushed. Thus, INET heap memory usage was huge, especially when a lot of transmissions occurred.


levy:
Thank you for submitting this pull request. I will review and merge it this week.

levy:
I have committed a different solution into the integration branch that fixes the indefinite heap growth without affecting fingerprints.

Once again thanks for reporting this bug!

riebl:
Thank you for fixing this issue properly!

PR #46, opened 2015-02-18
initial value of ssthresh

scimbe:
Two points:

1) the initial value of  ssthresh = 65535 is usually too small. Therefore the slow start is limited too early,
 As long no limit (like send buffer size) is given the sky should be the limit.

rfc5681: The initial value of ssthresh SHOULD be set arbitrarily high (e.g., to the size of the largest possible advertised window)

2) Netperfmeter uses a kind of socketcall to manipulate the max send buffer (sendQueue size) by a dirty workaround ( tcpCommand->getUserId()).
However, after the send buffer is set, the state of the connection should be reinitialized (e.g. to set the initial value of the ssthresh)

### DIFF
--- a/src/transport/tcp/TCPConnection.h
+++ b/src/transport/tcp/TCPConnection.h
@@ -152,7 +152,7 @@ class INET_API TCPStateVariables : public cObject
     TCPStateVariables();
     virtual std::string info() const;
     virtual std::string detailedInfo() const;
-    virtual void reInitialState() {};
+    virtual void setSendQueueLimit(uint32 newLimit);
   public:
     bool active;         // set if the connection was initiated by an active open
     bool fork;           // if passive and in LISTEN: whether to fork on an incoming connection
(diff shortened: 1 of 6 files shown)


src/transport/tcp/TCPConnection.h
```diff
@@ -152,6 +152,7 @@ class INET_API TCPStateVariables : public cObject
     TCPStateVariables();
     virtual std::string info() const;
     virtual std::string detailedInfo() const;
+    virtual void reInitialState() {};
```
ZoltanBojthe: Would be better write a `virtual void setSendQueueLimit(uint32 newLimit) { sendQueueLimit = newLimit; }` function,
use it in `TCPConnection::process_QUEUE_BYTES_LIMIT()`,
and overwrite `setSendQueueLimit` in `TCPTahoeRenoFamilyStateVariables` class for update `sstresh`, too.

scimbe:
Hope I hit it...

rhornig:
This seems to be ok from our side. @tuexen could you give a final nod before merging?

tuexen:
Setting the ssthresh to 0xffffffff makes sense.

rhornig:
Squashed the commits and merged into the master. You can delete now your own topic branch now (or rebase on master)

rhornig:
@scimbe Probably it would be good to include these changes also in the 'integration' branch. Changes in master will not be automatically merged into integration.

PR #47, opened 2015-02-18
OLIA for CMT-SCTP

scimbe:
OLIA for CMT-SCTP.

OLIA is beside LIA a basic CMT-SCTP Congestion Control.
No test cases yet, but the nornet/ripe project asked for this.
https://www.nntb.no/nornet-cooperates-with-ripe-ncc/

### DIFF
--- a/.cproject
+++ b/.cproject
@@ -1,20 +1,18 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<?fileVersion 4.0.0?>
-
-<cproject storage_type_id="org.eclipse.cdt.core.XmlProjectDescriptionStorage">
+<?fileVersion 4.0.0?><cproject storage_type_id="org.eclipse.cdt.core.XmlProjectDescriptionStorage">
 	<storageModule moduleId="org.eclipse.cdt.core.settings">
 		<cconfiguration id="org.omnetpp.cdt.gnu.config.debug.1590007136">
 			<storageModule buildSystemId="org.eclipse.cdt.managedbuilder.core.configurationDataProvider" id="org.omnetpp.cdt.gnu.config.debug.1590007136" moduleId="org.eclipse.cdt.core.settings" name="gcc-debug">
 				<externalSettings/>
 				<extensions>
-					<extension id="org.eclipse.cdt.core.MachO64" point="org.eclipse.cdt.core.BinaryParser"/>
-					<extension id="org.eclipse.cdt.core.ELF" point="org.eclipse.cdt.core.BinaryParser"/>
-					<extension id="org.eclipse.cdt.core.PE" point="org.eclipse.cdt.core.BinaryParser"/>
 					<extension id="org.eclipse.cdt.core.GmakeErrorParser" point="org.eclipse.cdt.core.ErrorParser"/>
 					<extension id="org.eclipse.cdt.core.CWDLocator" point="org.eclipse.cdt.core.ErrorParser"/>
 					<extension id="org.eclipse.cdt.core.GCCErrorParser" point="org.eclipse.cdt.core.ErrorParser"/>
 					<extension id="org.eclipse.cdt.core.GASErrorParser" point="org.eclipse.cdt.core.ErrorParser"/>
 					<extension id="org.eclipse.cdt.core.GLDErrorParser" point="org.eclipse.cdt.core.ErrorParser"/>
+					<extension id="org.eclipse.cdt.core.MachO64" point="org.eclipse.cdt.core.BinaryParser"/>
+					<extension id="org.eclipse.cdt.core.ELF" point="org.eclipse.cdt.core.BinaryParser"/>
+					<extension id="org.eclipse.cdt.core.PE" point="org.eclipse.cdt.core.BinaryParser"/>
 				</extensions>
 			</storageModule>
 			<storageModule moduleId="cdtBuildSystem" version="4.0.0">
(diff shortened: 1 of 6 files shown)


rhornig:
Thanks for the pull request. We had a small technical review:

- Could you change all identifiers to camel case? Underscores can be used if they are really justified (like the variable is named the same way in the standard).
- Why do we need both "olia-like" and "like-olia" for cmtCCVariant? We prefer to have one unambiguous name only. 
- the new value for the cmtCCVariant parameter should be documented in NED.
- Where is OLIA specified in detail? A reference to the algorithm's description would be great in the NED file, so a reader or later maintainer can verify the formulae.
- We need some documentation in the module's NED file that it supports OLIA.
- If a variable is OLIA specific it's name should contain "olia" so we could have an easier time later if we want to refactor the class. (i.e. are  assoc_best_paths; assoc_max_w_paths; assoc_collected_paths; OLIA specific variables?)
- in updateOLIA(uint32 w, const uint32 s, const uint32 totalW, double a, const uint32 mtu, const uint32 ackedBytes, SCTPPathVariables* path); it would be great to have more descriptive parameter names (it's hard to know what is w, s and totalW)
- an example would be great to show the algorithm in action (and the example's fingerprint could be used in the fingerprints tests: test/fingerprint/examples.csv) - this one would be ok at a later stage in a separate pull request.

tuexen:
Hi Martin,

could you provide a reference to a paper/internet-draft/rfc/whatever describing the algorithm you implemented? That way a reviewer can try to make sure that the code does what the document describes.

Best regards
Michael

scimbe:
I will work on this the day after tomorrow.  I will add a link to the ietf draft in the Ned file and will post it here.  Sry,  I forgot to commit the Ned file for this push request. Unfortunately I am on the Airport without access to the code... 

I will overwork the names, too and come back with an update..

scimbe:
> Could you change all identifiers to camel case?
1) Yes, but I m not sure if it is a good idea for  the calculation. I try both variants and commit the variant that is more easy to understand (on base of the draft). 

> Why do we need both "olia-like" and "like-olia" for cmtCCVariant? We prefer to have one unambiguous name only.

2) I did it like for mptcp-like and like-mptcp, which is still used in the code. However, I will change it to whatever you like.  

> the new value for the cmtCCVariant parameter should be documented in NED.
3) Agreed. I forgot to add the SCTP.ned file in the commit. 
Should I add the info for mptcp-like and cmtrpv1, too? 

> Where is OLIA specified in detail? 
4) https://tools.ietf.org/html/draft-khalili-mptcp-congestion-control-05

> A reference to the algorithm's description would be great in the NED file, so a reader or later maintainer can verify the formulae.
5) OK. I will add it. 

> We need some documentation in the module's NED file that it supports OLIA.
> If a variable is OLIA specific it's name should contain "olia" so we could have an easier time later if we > want to refactor the class. (i.e. are assoc_best_paths; assoc_max_w_paths; assoc_collected_paths; > OLIA specific variables?)

6) Mmmmh. Currently the variable cmtCCVariant is used to switch the CMT CC. 
I don't think to change this in the current code state seems a good idea. There is no factory pattern or something like that to control this by name. Perhaps during refactoring this could be a way, but on the current code base I don't see a reason to rename this variable. 

> in updateOLIA(uint32 w, const uint32 s, const uint32 totalW, double a, const uint32 mtu, const uint32 ackedBytes, SCTPPathVariables* path); it would be great to have more descriptive parameter names (it's hard to know what is w, s and totalW)

7) I see. I will add some further infos in the declaration. 

> an example would be great to show the algorithm in action (and the example's fingerprint could be 
> used in the fingerprints tests: test/fingerprint/examples.csv) - this one would be ok at a later stage in a > separate pull request.

8) As I said. I will come up with this later... If you want we can create here a ticket in the bug tracker that is assigned to me... Currently I am more interested to commit further TCP code (mptcp).... I only committed the OLIA code on the request of the Rite/Nornet Project.

rhornig:
> 2) I did it like for mptcp-like and like-mptcp, which are still used in the code. However, I will change it to whatever you like.

Multiple values that mean the same thing can be confusing for users. We should decide which value we like better and stick to that.

> 3) Agreed. I forgot to add the SCTP.ned file in the commit. 

Should I add the info for mptcp-like and cmtrpv1, too?
Yes please. Any extra documentation is welcome.

> 6) Perhaps during refactoring this could be a way, but on the current code base I don't see a reason to rename this variable.

Ok.

> 8)  If you want we can create here a ticket in the bug tracker that is assigned to me...

That would be a solution, but we are not comfortable merging code that is not exercised at all in any test or example. Even a simple example that executes the above code would be great as it could be added as a fingerprint/smoke test. We would rather wait with the merge until at least a minimal example exists. Rite/Nornet project could take over the code from this pull request directly.

I agree on all your other points and thanks a lot for the update.

dreibh:
Actually, the project using the OLIA patch is RITE, not RIPE. RIPE is a different project.
RITE project: http://riteproject.eu/ .

dreibh:
I could extend the NetPerfMeter example with an OLIA example. In fact, it would set cmtCCVariant="olia".

There is some legacy naming in the CCs. In fact, "MPTCP-llke" or "Like-MPTCP" should just be called "lia", since this is the name of the algorithm. Originally, it was supposed to be the standard algorithm for MPTCP, and the only one existing at time of writing the LIA CC code.

CMT/RPv1 and CMT/RPv2 are different algorithms, so it makes sense to keep them separate.

rhornig:
Thanks for the clarification @dreibh . @scimbe could you make the discussed modifications? Either by adding a new commit to the pull request or rewriting the original commit? 

We are planning of the release of INET 2.6 and would be great if the patch could make into the release.

scimbe:
1) I changed most things to camel case, even I kept the "_" notation in the calculation (like in the draft).
2) Changed it to "lia" and "olia"
3,4,5,6) Added more information in SCTP.ned, incl. the link for the OLIA draft
7) done
8) @dreibh can you change the netperfmeter example to LIA and OLIA
If you want use my branch, you can find it here: https://github.com/scimbe/inet-push/tree/feature/martin/olia_lia_sctp (it is public)

rhornig:
Thanks a lot. Everything looks fine. I have merged this into master as c551751. Please note that I had to squash the commits into one and remove some files as some unrelated changes in .oppbuildspec and the .project file crept in. Please delete your local feature branch to avoid confusion in the fututre.

PR #48, opened 2015-03-03
Topic/netperfmeter with traceplayer

dreibh:
Added a trace player feature to NetPerfMeter:
A trace is an ASCII table of the following format:
<time> <frame size> [<stream ID>]
After the given time, a frame of the given size is generated in the given stream (default: 0; only SCTP supports different stream numbers).

Upstream has never merged NetPerfMeter into StandardHost. Moved NetPerfMeter configuration to new node module NetPerfMeterHost => no need to modify core INET files any more.

Modified the NetPerfMeter example to use Router and NetPerfMeterHost instead of StandardHost.


rhornig:
Code style looks OK. Example runs and code builds with the NetPerfMeter feature turned off.

dreibh:
Description of NetPerfMeter:

The NetPerfMeter application (https://www.uni-due.de/~be0001/netperfmeter/) is a network performance meter for the UDP, TCP, MPTCP, SCTP and DCCP transport protocols over IPv4 and IPv6. It simultaneously transmits bidirectional flows to an endpoint and measures the resulting flow bandwidths and QoS. Since the application behaviour is the same for all underlying Transport Layer protocols, it allows to compare the Transport Protocol performances (e.g. TCP vs. SCTP). NetPerfMeter is a multi-platform software that currently works under Linux, FreeBSD and MacOS X. The NetPerfMeter model simulates the NetPerfMeter application, for comparing Transport Protocol performances in simulations. Since the parametrisation and features are similar to the real-world application, the model also makes comparisons of simulation results and real-world measurements easy.

avarga:
Thanks, great! Docu added on integration, after some wording changes (the text started with the description of the real-life NetPerfMeter app and not that of the model, this had to be changed).

PR #51, opened 2015-03-06
Added the version define that is already present in INET 2.6.0

tillsteinbach:
This is almost obsolete with the merge! But can we set the INET_VERSION to 0299?


rhornig:
The master got already merged so the version number is now available in integration too. As this is a hexadecimal number, the next version will be 2.100 which is 0x0264 in hex version number format.

PR #52, opened 2015-03-06
compat.h: check_and_cast should be defined in optional omnetpp namespace not in inet!

tillsteinbach:
Please check if it is ok to use NAMESPACE macros in inet.


rhornig:
NAMESPACE macros are fine.  They are part of the public API and were introduced back in 2008 so all omnet 4.x versions have them.

PR #57, opened 2015-03-11
Add data receive handling via indications. Add parameter windowScalingFa...

ruengeler:
...ctor

### DIFF
--- a/src/inet/transportlayer/contract/tcp/TCPCommand.msg
+++ b/src/inet/transportlayer/contract/tcp/TCPCommand.msg
@@ -39,6 +39,7 @@ enum TcpCommandCode
     TCP_C_ABORT = 6;         // abort connection
     TCP_C_STATUS = 7;        // request status info (TCP_I_STATUS) from TCP
     TCP_C_QUEUE_BYTES_LIMIT = 8; // set send queue limit (in bytes)
+    TCP_C_REQUEST = 9;       // send request to TCP to deliver data
 }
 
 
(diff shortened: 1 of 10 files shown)


avarga:
There seem to be two unrelated changes in this patch, could we tackle them in separate pull requests?
   - TCPSocker::setState(enum State state) -- I'd drop this, as it causes leaky abstraction (state is managed internally, it's should not be possible to overwrite it from outside); and it's also unused.
   - int windowScalingFactor parameter in TCP.ned -- it's unrelated, let's do it in a separate pull requests. Also, it would be better not to read the NED parameter read every time, but read once in the initialize() of TCP, and stored in a class variable.

avarga:
TCP_I_DATA_NOTIFICATION + TCP_C_REQUEST seems like a limited version of the real-life socket read syscall. Limited in the sense that only data that have already arrived can be read (one cannot request data in advance), and one cannot specify an amount (all data are sent up).

We may want to implement the full functionality of socket read sometime (there's been one attempt already, see attic/tcp_socket_read), so I suggest that TCP_C_REQUEST be renamed to TCP_C_READ. Later we can extend it, e.g. to allow queueing read requests, and to be able to specify the number of bytes to read in a control info object.

@tuexen, @scimbe, @ThomasReschka: what do you think?

rhornig:
Ups. Sorry we missed your patch fixing the issues. Thanks  a lot it looks good. I have merged the two commits into a single commit and re-based on top of integration: 3e140b35394dc7084277db85edfec40164c861f0

It's now merged. Please delete your original branch containing the pull request.

PR #58, opened 2015-03-13
networklayer/configurator/ipv4: multi-homed network auto-configurator for integration

dreibh:
The patch adds the multi-homed auto-configurator capability to the IPv4 network auto-configurator. When assigning a non-zero property "netID" (network ID) to links, the corresponding links and interfaces belong to the given network. For each network ID, the routing is computed separately. netID==0 means that the links (and corresponding interfaces) are shared with all paths.

Use case: mutli-path transport with CMT-SCTP
Example: 2 endpoints are connected via two different networks, for example a wired network and a wireless network. With the original auto-configurator, all routing between the two endpoints take the single, shortest path. With the updated auto-configurator, the links of both networks get different network IDs. Then, for each network ID (e.g. 1=wired, 2=wireless), separate routing tables are computed. Traffic send into the wired network then uses the wired network (ID 1), while traffic sent into the wireless network uses the wireless network (ID 2). To model a shared bottleneck link, the link would get ID 0 assigned.


dreibh:
Note, that it should be straightforward to extend this feature to IPv6 as well, once it is merged for IPv4.

avarga:
Thanks for the patches!

A quick question: do fingerprint tests still pass after this change? (`cd tests/fingerprint; ./fingerprints`)

The original author of IPv4NetworkConfigurator is @levy, it would be a good idea for him to have a look at the changes.

Some remark that'll become important later: The code style in some commits look a bit different from the usual INET code style (see http://inet.omnetpp.org/CodingConventions.html, recently updated). To resolve this, we could do the style changes after the code is merged (after it passed review etc), or you could do it in this pull request...

dreibh:
I have made two fixes and merged the latest "integration" changes. Now, the fingerprint tests are passed.

The source code formatting has been adapted as well.

levy:
Thank you for submitting this pull request. I reviewed your changes, and I have a few concerns regarding the added feature and its implementation. I understand that you want to configure more complex scenarios that is impossible using the current IPv4NetworkConfigurator.

I don't really like the idea of having a NED channel parameter called netID. The configurator already takes a complex XML parameter to control its behavior, and it seems to be an unnecessarily complication to also take parameters related to network configuration from NED. In addition, in your original example you mentioned a wired and a wireless network, I don't know how am I supposed to configure the netID for a wireless network or interface (without a channel)?

I'm not sure whether extracting the networks, computing the configurations, and performing them in a loop for all "virtual" networks is a good idea in general. Also, the introduction of the netID="0" network that includes all interfaces and links, seems to me quite accidental. What if an interface should be included in 2 out of 3 topologies?

dreibh:
Adding netID to a channel has been an easy way to add the new parameter without requiring many changes. Of course, if there is no channel, there is no netID (in the scenarios I used, I always had a channel). May be, a netID parameter could be added to the interfaces instead? What kind of approach do you suggest?

netID=0 is an easy way to specify "use in all topologies". This is already sufficient for many topologies. Of course, having a list of netIDs would add more flexibility. It would be quite easy to do this (just a few lines of code), so, once we can agree on where to specify the netID, I can make it a string and parse a vector of netIDs from the string.

dreibh:
Maybe INic is a good place for the nedID?

levy:
Sorry for the late answer but I was ill last week.

I suggest to avoid using node/link/interface NED parameters related to the network configuration. I think we should rather change the XML configuration file format. The XML configuration could be used to filter the topology, the considered destinations, the affected routing tables, etc. We could extend the configuration with function like XML elements that would instruct the configurator to execute certain algorithms (e.g. the static route calculator). In addition to solving the original issue this would allow having more flexible configurations. I will figure out how the XML configuration should look and come back.

levy:
How about adding an autoroute XML element that specifies the following:
 - the set of affected nodes where the routing tables will be updated
 - the set of destination interfaces for which the shortest path will be computed
 - the set of rules that determines the cost of nodes/links in the topology when computing the shortest paths

If there's no autoroute element in the configuration, then one is assumed by default. The default autoroute updates all routing tables with routes towards all interfaces in the network using the default metric.

    <autoroute sourceHosts="*" destinationInterfaces="*"/>

In general, the autoroute element would look something like:

    <autoroute sourceHosts="*" destinationInterfaces="*">
      <node hosts="*" cost="1">
      <link hosts="*" interfaces="*" cost="1">
    </autoroute>

where the sourceHosts attribute specifies a list of host full path patterns, the destinationInterfaces attribute specifies a list of interface full path patterns (including host full path). The node/link elements (several of them) specify the cost constraints for the shortest path algorithm (numbers, infinity, data rate based, delay based, error rate based, etc.).

Unfortunately in your example, you would have to add as many autoroute elements as the number of networks. So, what do you think? Does it sound ok for your example?

levy:
I've pushed the IPv4NetworkConfigurator autoroute changes into the integration branch. I think this feature allows you to configure your network without further changes to the configurator.

Let me know if this isn't sufficient.

PR #107, opened 2015-03-16
CC function initialisation bugfix: need to initialise function to call before SACK

dreibh:
ccFunctions.ccUpdateBeforeSack needs to be initialised with &SCTPAssociation::cwndUpdateBeforeSack. Otherwise, ccFunctions.ccUpdateBeforeSack is nullptr, and the call to this function will fail.


rhornig:
Looks good.

PR #108, opened 2015-03-17
Set the correct size for resources in random browsing mode.

Tey:
The size of resources sent when using random mode (i.e., non-scripted mode) does not match the size configured by the user. This commit fix that.


rhornig:
Looks ok.

PR #110, opened 2015-03-17
Fix parsing and units in RTPAVProfilePayload32Sender::sendPacket().

Tey:
This commit fixes multiple things:
* the parsing of GDF files which only detects space characters as correct fields separators. All the example GDF files use tabulations as separators, so the parsing does not work correctly even for official examples (the type of video frames is wrongly parsed).
* there's a mistake between bits and bytes when computing the size of MPEG packets. The quantities in GDF files are in bits, but they are converted to bytes before the splitting loop.
* the payload size of MPEG packets is never set while the RTP receiver needs it.

With this commit, the RTP examples work correctly now...


rhornig:
Looks good.

PR #111, opened 2015-03-17
Fix a memory leak when handling undeliverable UDP packets.

Tey:
The control info associated with UDP packet is removed from it when receiving an undeliverable packet. It and the packet is then given to `UDP::processUndeliverablePacket()`, but this method only releases the packet (which does not release the control info since it is not associated with the packet anymore). The calls to `sendErrorMessage()` (icmp/icmpv6) do not release the control info either, so there's a memory leak.

I don't know if the control info should be released in `UDP::processUndeliverablePacket()` or in `sendErrorMessage()`, but it must be released anyway. I choose the former for this commit.

I'm really surprised that no one has ever found that leak in the past as OMNet++ runs out of memory quickly when using UDP broadcast/multicast streaming, which does not sound like an uncommon scenario.


rhornig:
Looks good.

PR #112, opened 2015-03-17
Allow initial positions to contain non-integer values.

Tey:
This commit allows the initial positions of nodes (set in omnetpp.ini) to contain non-integer values, and to show the exact positions of nodes in the UI instead of rounding them to nearest integer positions. The UI is already able to handle non-integer positions so why would the mobility module try to feed it with only integer positions?

By integer positions I mean position where `x` and `y` are integers.

Please, consider merging this commit in inet.


rhornig:
Looks good.

PR #113, opened 2015-03-17
Allow the user to disable RTP logging to file.

Tey:
The RTP log files are generated automatically by the `RTPPayloadReceiver` modules, 2 for each nodes. When simulating lots of nodes, you end up with lots of files even if you are not interested by their content.

This commit permits to disable the generation of these files based on the user configuration.

Please consider merging it into the inet framework.

### DIFF
--- a/src/transport/rtp/RTPPayloadReceiver.cc
+++ b/src/transport/rtp/RTPPayloadReceiver.cc
@@ -38,7 +38,8 @@ void RTPPayloadReceiver::initialize()
 {
     const char *fileName = par("outputFileName");
     const char *logFileName = par("outputLogFileName");
-    if (strcmp(fileName, "")) openOutputFile(fileName);
+    if (strcmp(fileName, ""))
+        openOutputFile(fileName);
     if (strcmp(logFileName, ""))
     {
         char logName[200];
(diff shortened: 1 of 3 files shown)


Tey:
Done. Note that writing to or closing a closed stream does not seem to throw any exceptions on Windows or Linux/Ubuntu.

avarga:
Yes, throwing exceptions needs to be explicitly enabled on stream objects, by default it only sets error bits if something goes wrong. Thanks for adding the guards conditions! We think it makes the code easier to read, by making it more obvious that logging may be turned off. Also spares some cycles in some cases.

avarga:
Hello Tey,
You probably noticed the recent update of the INET web site, and that we'd like to expand community participation in the project. We have established a network of Component Advisors as a way of bringing in expertise and giving core users/contributors more influence and control over the project (http://inet.omnetpp.org/ComponentAdvisors.html). Given your recent commits and apparent interest in the RTP model in INET, I'd like to ask whether you'd be interested in becoming a Component Expert for RTP. Your level of involvement would be entirely up to you, and I could add you with the github nick to the Advisors table instead of your full name if you wish. Please let me know :)

Tey:
Hi Andras,
Thanks for the offer, but I've only installed OMNet++/INET some weeks ago to do quick tests and I probably won't use it again as I now have nothing more to test.

avarga:
No problem, thank you for your contributions so far, I very much appreciate the efforts you've put in!

PR #116, opened 2015-03-19
Look up addresses in Init-Ack chunk to verify source address

ruengeler:


rhornig:
Looks ok. Thanks for the rename.

PR #118, opened 2015-03-20
Immediate scheduler

arturscussel:
It's already without the defines. Thanks

### DIFF
--- a/src/linklayer/ext/cSocketRTScheduler.cc
+++ b/src/linklayer/ext/cSocketRTScheduler.cc
@@ -31,7 +31,6 @@
 
 #define PCAP_SNAPLEN 65536 /* capture all data packets with up to pcap_snaplen bytes */
 #define PCAP_TIMEOUT 10    /* Timeout in ms */
-#define PCAP_IMMEDIATE 1   /*Set the pcap immediate mode*/
 
 #ifdef HAVE_PCAP
 std::vector<cModule *>cSocketRTScheduler::modules;
@@ -71,7 +70,6 @@ void cSocketRTScheduler::startRun()
     if (setsockopt(fd, IPPROTO_IP, IP_HDRINCL, (char *)&on, sizeof(on)) < 0)
         throw cRuntimeError("cSocketRTScheduler: couldn't set sockopt for raw socket");
 #endif
-
 }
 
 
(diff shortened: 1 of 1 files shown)


rhornig:
Hello @tuexen @ruengeler ,
we have received a pull request that turns on the immediate mode for pcap. As the documentations suggest, the scheduler was not running in immediate mode until now. According to the docs http://www.manpages.info/macosx/pcap.3.html the read operation waits for several packets to arrive until the timeout expires so we are not seeing the packets as they arrive.

Would you confirm that it's OK to use the immediate mode? It seems justified IMHO for something that's called "realtime", but you may have some other experiences. Please give an OK, if we can merge.

src/linklayer/ext/cSocketRTScheduler.cc
```diff
@@ -117,11 +117,18 @@ void cSocketRTScheduler::setInterfaceModule(cModule *mod, const char *dev, const
 
     /* get pcap handle */
     memset(&errbuf, 0, sizeof(errbuf));
-    if ((pd = pcap_open_live(dev, PCAP_SNAPLEN, 0, PCAP_TIMEOUT, errbuf)) == NULL)
-        throw cRuntimeError("cSocketRTScheduler::setInterfaceModule(): Cannot open pcap device, error = %s", errbuf);
+    if((pd = pcap_create(dev,errbuf)) == NULL)
```
tuexen: Use 
pcap_create(dev, errbuf)

src/linklayer/ext/cSocketRTScheduler.cc
```diff
@@ -117,11 +117,18 @@ void cSocketRTScheduler::setInterfaceModule(cModule *mod, const char *dev, const
...
+        throw cRuntimeError("cSocketRTScheduler::setInterfaceModule(): Cannot create pcap device, error = %s", errbuf);
     else if (strlen(errbuf) > 0)
         EV << "cSocketRTScheduler::setInterfaceModule(): pcap_open_live returned warning: " << errbuf << "\n";
 
+    /* apply the immediate mode to pcap */
+    if(pcap_set_immediate_mode(pd,1) != 0)
```
tuexen: Use
pcap_set_immediate_mode(pd, 1)

src/linklayer/ext/cSocketRTScheduler.cc
```diff
@@ -220,7 +227,9 @@ bool cSocketRTScheduler::receiveWithTimeout()
 
     found = false;
     timeout.tv_sec = 0;
-    timeout.tv_usec = PCAP_TIMEOUT * 1000;
+    timeout.tv_usec = 100;
```
tuexen: Maybe it is better to use some #define the the timeout. Then is it easier to change parameters, if needed.

tuexen:
The changes with my editorial changes look good.
However, I'm not sure if the functions newly called are available in the pcap implementations used on the platforms supported by inet. If that is true, the patch should go in.

rhornig:
Thanks, I will check whether the pull request builds correctly on all platforms.

avarga:
Found an issue in cSocketRTScheduler, see #120

rhornig:
Reformatted a bit and made the timeout is now defined with a macro. Merged into master as 81596a.

Issue #120, opened 2015-03-23
cSocketRTScheduler logic error


avarga:
cSocketRTScheduler causes the deadline of internal simulation events to be missed.

The `cSocketRTScheduler::receiveUntil(const timeval& targetTime)` function is supposed to wait until `targetTime` or until the first packet arrives, whichever is sooner. (`targetTime` is the time at which the next OMNeT++ simulation event is supposed to be executed.) However, internally it relies on repeatedly calling the parameterless `receiveWithTimeout()` function until `targetTime` arrives.  Since `receiveWithTimeout()` doesn't know `targetTime` (it uses a constant timeout, `PCAP_TIMEOUT`), it will miss `targetTime`.

`receiveWithTimeout()` should observe `targetTime`. This will allow a larger `PCAP_TIMEOUT` now that pcap immediate mode is used. `PCAP_TIMEOUT` then will essentially become the user interface refresh interval.

rhornig:
The scheduler now correctly computes the required timeout value. On Linux the UI refresh time was set to 100ms. On other platforms the wait time is still correctly calculated, but the scheduler needs to poll the interface regularly to because of missing APIs that allow waiting without blocking. On non Linux platforms the poll time was set to 500us.

PR #122, opened 2015-03-25
emulation: fix for #120

rhornig:


rhornig:
@arturscussel Check this change as it may improve the timing of the scheduler and at the same time reduce the CPU usage.

arturscussel:
No, it's very slow, ping with 70 ms, in express mode! I had not understood the #120 , but now I think I get it, I will give a look on that.

rhornig:
Yes, I'm testing it too. I see the slowdown too.

arturscussel:
Yeah, the problem with that it's because now select() is blocking the code, doesn't matter if the pcap is in the immediate mode. When it is waiting the next message, it always enters on receiveWithTimeout(100000), because the targetTime is infinite (it is equal to LONG_MAX). So it's blocking it with the 100 ms timeout.

arturscussel:
The time for processing the message it's less than 300 us, if you put receiveWithTimeout(500), in the worst case the ping will be like 800 us. And you can just alter the while header to something as while(diff_sec<1 || diff_usec<1000). The cpu usage is ok.

rhornig:
Yeah, the fix actually is good, it's just at the same time I have increased the timeout back to 100ms because I was assuming that the select will return immediately when a packet is received. I'm not really experienced in this stuff, but why is the select blocking for 100ms if the immediate mode is set? I have assumed that that's what setting immediate does.

arturscussel:
There was two time outs, the select's one and the pcap's one. If you see the select() parameters, you will see that it's just watching the timeout, the other thing are NULL, so it's just like a sleep. The pcap timeout doesn't exist anymore with the pcap immediate, when I call pcap_dispatch it will only check if there is a packet, if not move on.

arturscussel:
The good side of using LINUX it's because if you look inside the #ifdef LINUX, you will see a different select(), the first parameter looks for a file descriptor associated to that pcap, so when something happens there, it leaves and you can use the receiveWithTimeout(100000), because it will leave immediately. But there is a #ifdef LINUX, so I don't know if that works for other OS, we can take it off and test is.

arturscussel:
no, there is a function not available for windows, pcap_get_selectable_fd()

rhornig:
I'm testing on Linux and I was wondering why the blocking happens and it turns out:
There is no such macro defined as LINUX it is called __linux__ :) which means the actual code intended for linux never ran. I have corrected this and in express mode I'm getting response times on pair with the netive interface while the CPU utilization is below 1%...

I will check check whether that function is available on Mac. If nothing else, we can have a smaller timeout on windows (and live with the higher CPU utilization) and have a longer timeout on linux.

Thanks for helping this, 
@arturscussel have a cookie @changetip

changetip:
Hi @arturscussel, @rhornig sent you a Bitcoin tip worth a cookie (6,080 bits/$1.50), and I'm here to deliver it ➔ **[collect your tip](https://www.changetip.com/collect/663748)**.


**[Learn more about ChangeTip](https://www.changetip.com/tip-online/github)**

arturscussel:
Ahhh I was wondering why it was not entering in the #ifdef, so I just put a define LINUX 1 to test if the code worked haha, quick and dirty way, thank you for the tip.

rhornig:
Pushed a new commit. Please test it. It should now work on Linux. On other opsystems the polling tme is set to 500us which should be low enough. Please test it.

arturscussel:
ok, here is everything working fine. Thanks

PR #127, opened 2015-04-10
Fix: removed debug code that caused unwanted data transmissions.

dreibh:
Some debug code got merged that set frameSize=1424 when frameSize=0 (i.e. no transmission) was configured => unwanted data transmissions. Fixed this by removing the debug code.


rhornig:
Looks ok.

PR #128, opened 2015-04-10
Added target "makefiles-static" to Makefile => easy static builds

dreibh:
I frequently do static builds instead of shared object builds. Therefore, I just moved the opp_makemake options into a separate options line and added another build target "makefiles-static" for the static build. So, creating a static build can be done with "make makefiles-static && make". The old behaviour as default remains unchanged (i.e. shared-object builds with "make makefiles && make").


rhornig:
Looks great. Thanks.

PR #133, opened 2015-04-10
Bugfix: Also send SCTP_I_SENDQUEUE_FULL indication for unordered messages

dreibh:
The queue size limit was only checked for ordered messages. Consequently, no limitation was done for unordered ones.


rhornig:
Thanks Thomas merged now. 

Just a note for future commits: when you add commit messages, could you add a prefix to the message that specifies what component actually the commit is touching? i.e: 

sctp: description of commit
tcp: description of commit
netperfmeter: description of commit
etc.

We are trying to do this because it is useful when you check only a git log. It is rather hard to identify the context from the log files without this information.

PR #135, opened 2015-04-16
sctp: added default parameter settings to connect() call of SCTPSocket

dreibh:
sctp: added default parameter settings to connect() call of SCTPSocket. Then, in the usual case, a connect() just takes address and port (like it is for TCP).


rhornig:
Looks good.

PR #138, opened 2015-04-17
Extended run_inet script with option to run Valgrind

dreibh:
Extended run_inet script with option "--valgrind" as first parameter. If given, the simulation (executable only) is started with Valgrind (http://valgrind.org/) using memcheck parameters => convenient for debugging issues like uninitialized variables and memory problems.

### DIFF
--- a/src/run_inet
+++ b/src/run_inet
@@ -1,8 +1,17 @@
 #! /bin/sh
 DIR=`dirname $0`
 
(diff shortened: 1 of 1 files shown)


rhornig:
Had to make some minor modifications + updated the WHATSNEW file. Merged as 0254257e5834da71acb5957c04e7ec548184404b

PR #141, opened 2015-04-17
Udp serialize

arturscussel:
I was having some problem transferring some UDP datagrams and I think it wasn't doing the checksum correctly, so I wrote this code and for now it's working fine, please give a look. Thanks.


avarga:
Thanks for the patch, we'll check and merge it soon! FYI, we just put a lot of work into improving serializers on the integration branch, have you seen it? Unfortunately, those improvements would be very hard to backport into master, because they also depend on changes in other parts of INET (i.e. outside the serializer code).

arturscussel:
No, I didn't see it, thanks, I will check it.

PR #144, opened 2015-04-23
Add Packetdrill for UDP and two test cases

ruengeler:
Basic files for packetdrill.
One Problem is that the parser.y and lexer.l have to be converted to parser.cc and lexer.cc before make makefiles can be called. The necessary commands are:
bison --output=src/inet/applications/tunapp/parser.cc --defines=src/inet/applications/tunapp/parser.h --report=state src/inet/applications/tunapp/parser.y
flex -o src/inet/applications/tunapp/lexer.cc src/inet/applications/tunapp/lexer.l
I do not know where to put these commands in the tool chain. I have always called them manually.


rhornig:
I will look into this how to set up the build process to automatically do this.

rhornig:
Thanks @ruengeler , got a bit further with the review:

- It seems that the PacketDrillApp and the PacketDrill host files are inconsistent. PacketDrillHost assumes that the PacketDrillApp has sctpIn,sctpOut, tcpIn and tcpOut gates, while in fact the PacketDrillApp has only these:

    gates:
        input udpIn @labels(UDPControlInfo/up);
        output udpOut @labels(UDPControlInfo/down);
        input tunIn @labels(TunInterface/up);
        output tunOut @labels(TunInterface/down);

- Additionally the PacketDrillHost should be moved to a separate inet.node.tun package from the inet.node.inet. (and the inet.node.tun package should be added to the TUN feature). the reason is that INET should be compilable with the TUN feature turned off. With TURN turned off this would give a NED error in the generic inet.node.inet package as TUN specific classes (PacketDrillApp etc. would be disabled).

rhornig:
Hello @ruengeler,

I'm picking up the pull request again. unfortunately I did not realized that you have commited something after my last commit, so the change went unnoticed. Please after new commits post also a comment probably mentioning my username so I will be notified about the changes you've made.

As for the changes: I made a new review and made some changes to fix some issues:
- rebased the whole PR on top of the INET 3 master branch and created a new topic branch for it (topic/packetdrill)
- moved the packetdrill app and related components to their own folder (inet.applications.packetdrill)
- added a new feature for packetdrill so it can be turned on/off
- removed defaults from the NED file that cannot be justified. 
- added a bash script that calls bison and flex to generate the parses if the .y and .l files change.

Please use the following branch as a base for further changes (do not commit on top of this PR). I would prefer that you close this PR, and submit a new PR based on the content of the topic/packetdrill branch.
https://github.com/inet-framework/inet/tree/topic/packetdrill

Outstanding issues:
- The code seems to have issues when compiled on Windows and Linux. Relying on the presence of HAVE_FMEMOPEN macro is not a good solution. Additionally there is no fmemopen() on Windows. The code must compile on all platform before the PR can be merged so we have to find some crossplatform replacement for this function. It's ok also to use C++ code here, we don't have to stuck with C. Probably something like istringstream() ??? http://www.cplusplus.com/reference/sstream/istringstream/

ruengeler:
@rhornig 
Hi Rudolf,
I closed the PR and will try to pick up the pieces on the packetdrill branch.

PR #146, opened 2015-04-24
Sctp bugs

ruengeler:


rhornig:
Hi @ruengeler , could you please prefix each commit messages with the component they are touching. In these i.e.

sctp: commit message

we are asking this because it makes browsing the git history much more easy as you can see the scope of the change immediately and you do not have to look into the actual commit. (i.e. seeing the message "Set the primary path when the path is established" without any context is not really helpful).

PR #150, opened 2015-05-12
httptools: Added support for SCTP to browser and server modules

dreibh:
httptools: Added support for SCTP to browser and server modules. Then, both modules can be used as SCTPApp. An example configuration has been added to examples/httptools/socket/pairsocket: scripted-with-sctp.


rhornig:
@dreibh Thanks, I've checked this PR. It seems to contain too much commits, some of them are unrelated to the actual feature (the PR should touch only HTTPTOOLS sources). Merge commits should be omitted, too. Also we have to ensure that the HTTPTOOLS feature can be used while the SCTP feature is turned off.

rhornig:
Thanks for the contribution. We are postponing this feature because we are in the process to re-architect how the StandardHost is organized. The new architecture will allow opening several sockets for Apps (including even multiple ones at the same time. i.e. the same app can open UDP,TCP and SCTP sockets at the same time). This change is scheduled for INET 4.

PR #151, opened 2015-05-13
httptools: bugfix

dreibh:
httptools: fixed registration of server when servers are inside other modules


rhornig:
Looks fine. Thanks for the fix.

PR #152, opened 2015-05-19
SCTPSocket bugfix for ordered/unordered mode selection

dreibh:
sctp: Bugfix: SCTPSocket::send() may send a message in unordered mode *only* if its kind is SCTP_C_SEND_UNORDERED. In any other case, use ordered mode as default.


rhornig:
Look ok.

PR #154, opened 2015-06-06
Mynemo

fayruzrahma:


rhornig:
Hello @fayruzrahma, thanks a lot for the contribution. I've made a very preliminary review... I think NEMO could be useful part of INET if we sort out some issues. 

- As we are heading towards INET 3.0 (we plan to release this week), it would be great to have this work rebased against the current integration branch (that will become master once we release 3.0). We do not plan to add any further features to the 2.x branch.
- Comments in code should be in English. The codebase contains non-english comments, especially where the already existing code is modified. Code blocks got commented out, but those changes MUST be justified and proved that it does not affect previously working behavior (IPv6NeighbourDiscovery.cc, xMIPv6.cc etc.)
- We will definitely need an example that shows NEMO in action, but the pull request should NOT modify existing examples. Could you prepare an example as a separate configuration  instead of modifying a current example?
- It would be great to have some tests for the main features of NEMO (see tests/module), otherwise we will never be sure whether the code still works correctly after a refactoring or mayor change.
- I cannot see the reason for the existence of the EthernetInterfaceOnlyForward calss. Somhow I don't feel that NEMO should require a modification in the linklayer.
- I saw several places in the code where direct interface names like "eth1" and "wlan0' were hardcoded into the code. Any new component must work generally in the INET framework and must support any number of ethernet and wireless cards so those parts of the code should be fixed.

This was just a quick check on the code. If we can get these sorted out, we can continue the review to get  the component into INET 3.1

rhornig:
BTW, if you decide to move to INET 3.0 I can give some assistance for the porting.

Issue #163, opened 2015-07-31
EtherMACFullDuplex - move getNextFrameFromQueue() to scheduleEndIFGPeriod


tillsteinbach:
Dear INET-Community,

I want to suggest a change in the EtherMACFullDuplex module: Currently, the next frame is already requested before the IFG has finished. This way the next frame idles for the time of one IFG in the MAC. The behavior is not wrong, but it makes the implementation of shaper-algorithms more complex as the shaper cannot expect that the frame is sent immediately. Thus I would suggest that the next frame is requested after the end of the IFG. This is also much more intuitive.

If there are no arguments against such an extension I can offer to implement those changes and provide a pull request.

ZoltanBojthe:
The correct title is:
EtherMACFullDuplex - move getNextFrameFromQueue() to handleEndIFGPeriod

Issue #162, opened 2015-07-31
EtherMac (interaction with shaper modules)


tillsteinbach:
Dear INET-Community,

I want to suggest some extensions in the EtherMac modules. Traffic shapers become more complex and some of the algorithms require to have more knowledge of what is going on in the MAC e.g. to calculate credits.
Thus I would suggest to add the following to the Ethernet MAC modules:
- public getters for the transmitState
- public getters for the receiveState
- signal for the transmitState
- signal for the receiveState

If there are no arguments against such an extension I can offer to implement those changes and provide a pull request.

Issue #160, opened 2015-07-09
EtherMACFullDuplex cannot send frames larger than 1518 byte (MAX_ETHERNET_FRAME_BYTES)


tillsteinbach:
I would like to propose to change the macro MAX_ETHERNET_FRAME_BYTES to at least 1522 as IEEE 802.3ac standard increased the maximum Ethernet frame size from 1518 bytes to 1522 bytes.
It might be even better to set it to 1526 as IEEE standard 802.1ad allows double-tagging.

Another possibility would be to drop the check for maximum frame size in the Ethernet MAC to allow for jumbo frames (although not in the IEEE 802.3 Ethernet standard).

comments appreciated!
Best regards
Till

PR #165, opened 2015-09-07
Fixing/New Feature #160, #162 and 163 (EtherMAC)

tillsteinbach:
As discussed with @ZoltanBojthe, here is the pull request.

It contains the changes for the issues: #160 , #162 and #163. If one of the changes is undesired you would have to git-cherry-pick.

**As there are changes in the statistics this might break fingerprints! So please have a careful look that I did not make a mistake!**

Best regards
Till


rhornig:
As these are fixes required to be merged into the 3.x branch, could we rebase this on top of the master branch (instead of integration which will be 4.0 sometimes in the far future). Also getting rid of the upstream-merge commits would be great as they generate a lot of noise in the history.

tillsteinbach:
Hi Rudolf,
I'll give it a try! I'm not an expert in rebasing things, thus I would suggest opening a new branch from master and cherry-picking my commits there! I'll let you know if that worked.

rhornig:
That solution is totally acceptable and I'm doing it also in these cases :)

tillsteinbach:
closed for new pull request #166 for master

PR #166, opened 2015-09-07
Fixing/New Feature #160, #162 and 163 (EtherMAC)

tillsteinbach:
New pull request for master

It contains the changes for the issues: #160 , #162 and #163. If one of the changes is undesired you would have to git-cherry-pick.

**As there are changes in the statistics this might break fingerprints! So please have a careful look that I did not make a mistake! The module tests (eth*.test), were ok**

Best regards
Till

### DIFF
--- a/src/inet/linklayer/ethernet/EtherMAC.cc
+++ b/src/inet/linklayer/ethernet/EtherMAC.cc
@@ -119,6 +119,7 @@ void EtherMAC::processConnectDisconnect()
         if (!duplexMode) {
             // start RX_RECONNECT_STATE
             receiveState = RX_RECONNECT_STATE;
+            emit(receiveStateSignal, RX_RECONNECT_STATE);
             simtime_t reconnectEndTime = simTime() + 8 * (MAX_ETHERNET_FRAME_BYTES + JAM_SIGNAL_BYTES) / curEtherDescr->txrate;
             endRxTimeList.clear();
             addReceptionInReconnectState(-1, reconnectEndTime);
@@ -374,6 +375,7 @@ void EtherMAC::processMsgFromNetwork(cPacket *msg)
 
         // set receive state and schedule end of reception
         receiveState = RX_COLLISION_STATE;
+        emit(receiveStateSignal, RX_COLLISION_STATE);
 
         addReception(endRxTime);
         delete msg;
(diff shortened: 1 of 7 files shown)


tillsteinbach:
@rhornig, @ZoltanBojthe
Is there anything more I can do to support you?

rhornig:
@tillsteinbach @ZoltanBojthe made some tests and this does not seem to be so straightforward for #163 . Some tests are failing now that were not expected to fail because of the change. Tomorrow we will investigate this.

tillsteinbach:
Are there tests beside the fingerprints that verify the Ethernet timing? I thought I have to run the module tests only?

tillsteinbach:
@ZoltanBojthe 
Could you please recheck? It looks like I messed up my branch while cherry-picking. I also added more asserts.

ZoltanBojthe:
@tillsteinbach, @rhornig
Here is my branch:
https://github.com/inet-framework/inet/tree/tillsteinbach-refactor/EtherMAC_for_3.x
(I built it with official omnetpp4.6.)
When fingerprint test running, 35 examples generate ASSERT:
 for example the `/examples/ethernet/lans/  -f duplexswitch.ini -c SwitchedDuplexLAN -r 0` at event #2128.

If full duplex mac has internal queue and module receives a packet from upper layer,  the `processFrameFromUpperLayer()` fill the `curTxFrame` if its empty. With your changes, if it happen in `WAIT_IFG_STATE`, then `getNextFrameFromQueue()` function gets next packet from queue and overwrites `curTxFrame` pointer at the end of IFG.

Other problem: if full duplex mac module has external queue, then `TX_IDLE_STATE` transmit state always appear between `WAIT_IFG_STATE` and `TRANSMITTING_STATE`.

tillsteinbach:
@ZoltanBojthe 
did you check with 5b6cb61772479d84721ecec43867b3997c4687ac ? It contains a check that the frame is only selected from the queue when the Interface is idle. 

Concerning the state changes: Yes, this includes a TX_IDLE phase with zero duration. I actually thought that this is not wrong (especially with an external queue) as it signals that the next frame can be sent.
If that is undesired, a way to prevent this behavior would be to ask the external queue if it has frames and only otherwise set it to idle, but this might have other drawbacks and I don't like that solution. Another (better) possibility would be to schedule the beginSendFrames(). This way the requested frame from the upper layer would be in the event queue before the mac tries to send the frame. I don't know if that is an elegant solution as it involves one more event.

PR #171, opened 2015-10-13
TCP: Configurable window scale factor

ruengeler:
@rhornig 
To make the window scale factor configurable helps to compare the simulation to TCP stacks where you can set the factor via sysctl.


rhornig:
@ruengeler I have merged this, but made some modifications in 568985e472771a99b450fdf15077651aea813725

Reason: you do not have to check for the presence of a parameter with hasPar() as we always assume that a NED file and the corresponding CC file is in sync. If not, then we need runtime errors to detect the problem.

PR #176, opened 2015-11-03
Added INET_API Macro to serializers

tillsteinbach:
This fixes the "undefined reference to vtable" problem under windows.

I'm not completely in the picture with all this INET_API thing, but it seems like it is recently often forgotten in new code. Is there a way we can prevent these problems?


rhornig:
Is this issue happening during the INET build or rather during the build of a component which is using INET?

tillsteinbach:
The second. We are using the serializer to export simulated traffic streams in other tools

rhornig:
That macro is used to mark the classes/methods that are considered external API and should be exported by the linker. I'm not sure how we could enforce their presence except with a more careful code review... Would love to have some automated suite for this, but that issue surfaces only when you actually want to link against the library.

PR #188, opened 2016-02-26
StaticEllipseMobility

fastriker:
Create StaticEllipseMobility which places all hosts on a ellipse based on an given area over constraintAreaMin and constraintAreaMax.
Construction is based on Steiner's generation of a conic. Therefore the number of hosts defined in numHosts must be a multiple of 4.


rhornig:
Thanks for the contribution. We are postponing this feature because we plan to re-architect the whole mobility handling. We will take this as a requirement for the redesigned component.

PR #189, opened 2016-02-29
Make a StationaryCircleMobility which place hosts on a circle.

fastriker:
Similar to StationaryEllipseMobility but based on the distance parameter.


rhornig:
Thanks for the contribution. We are postponing this feature because we plan to re-architect the whole mobility handling. We will take this as a requirement for the redesigned component.

PR #192, opened 2016-03-17
Packetdrill: 

ruengeler:
Add functionality to use setsockopt in packetdrill test


rhornig:
Thanks a lot for the PR. In the master branch we have previously merged your last PR (SCTP support for packet drill) which turned out to cause build errors on Windows so I regretfully had to revert the changes and re-release INET 3.2.3 without those changes. As this PR contains also those changes would you mind testing it on Windows? I can merge the change only if it cleanly compiles on all supported platforms (Linux, Mac, Windows) + both on OMNET 4.6 and 5.0RC. Later we will drop the OMNET 4.6 compatibility, but for now we must maintain it for a while.

Tip: if you are using iterators in the code you have to use the already deprecated it() syntax instead of the new *it. We need that for omnet 4.6 backward compatibility. (it will throw warning on omnetpp 5.0, but we have to live with that until we drop the omnet 4.6 compatibility in INET. At that point we will change all iterator references in one step).

Thanks, Rudolf

PR #198, opened 2016-04-21
Fixes for the EthernetSerializer

tillsteinbach:
Patch for problems mentioned in: https://groups.google.com/d/msg/inetframework-contrib/IHMo0JuI0Pw/z3WUHqc4CQAJ

- Fixed length field
- Fixed CRC
- Fixed padding

- Context feature to disable exception when serialized packet is shorter
than bytelength (Preamble and SFD not being serialized)
- Context feature to enable serialization of preamble and SFD


rhornig:
This issue has been fixed in the master using a different approach.

PR #202, opened 2016-06-06
More linktypes of libpcap to register serializers

tillsteinbach:
This is required for serialization of fieldbusses (CAN/Flexray) and raw Ethernet


rhornig:
@tillsteinbach After a bit of discussion, we have reverted this change. The extra enums are not used in INET but only in dependent projects. As these are anonymous enums it would be better to host them in the same project where they are actually used to avoid inconsistencies. (i.e. during a future cleanup we may accidentally clean these up if they are not used locally).

tillsteinbach:
Hi @rhornig,
I agree I can host these ENUMs in my own project(s). That way it also won't confuse INET users.

Would be great if C++ would allow extending ENUMs wouldn't it? OMNeT++ enums are also not extendable right?
:)

rhornig:
Well. At some point we may switch to scoped enums (enum class) from he current unscoped ones. At that point this PR would make sense.

PR #215, opened 2016-07-29
pathloss: add two-ray interference model

riebl:


levy:
Thank you for the pull request. Before merging, could you please tell me what's the relation between this model and the TwoRayGroundReflection model? I guess they are different.

riebl:
TwoRayGroundReflection is a simplification of TwoRayInterference when the distance between receiver and transmitter is large and perfect polarization and reflection can be assumed, see http://www.ccs-labs.org/bib/sommer2011using/sommer2011using.pdf

levy:
Thanks, merged.

PR #232, opened 2017-01-11
fixed ttl off-by-one error

juliusf:
Hi,

I was poking around with traceroute in a little simulation I'm currently playing with and found that routers in inet are forwarding IPv4 packets with ttl==0, which is not correct. It seems that this behavior has been accidentally introduced in de2b2a40ba6788c9d00be7f3265a011749a2fd51 

Best,

Julius


ZoltanBojthe:
The TTL decrementation also fixed: moved it to handleIncomingDatagram()

PR #233, opened 2017-01-11
fixed invalid TTL reuse for ICMP Echo Reply

juliusf:
Hi,

I found that when creating ICMP Echo Replys, inet would simply reuse the old IPv4ControlInfo from the request with swapped SRC and DST addresses.
This means that the packet would be sent with the TTL it has seen on the request packet. 
By setting it to 0, the IPv4 layer correctly resets it to the default TTL.

Best,

Julius


ZoltanBojthe:
Looks OK

Issue #250, opened 2017-05-01
Segmentation Fault in RadioMedium (dangling pointer)


riebl:
Occasional crashes occur in simulations with dynamic hosts, i.e. where radios are added and removed dynamically during simulation.
RadioMedium stores ongoing transmissions (ITransmission) and these have an getter for retrieving the transmitting radio (ITransmission::getTransmitter). However, with increasing number of network nodes the likelihood increases that a network host is removed while one of its transmission is still on its way. In this situation ITransmission::getTransmitter returns a dangling pointer.

A typical stack trace of such an error looks like this:
inet::physicallayer::RadioFrame::getListening() at RadioFrame.cc:60 0x7fffd3fdbb14	
inet::physicallayer::Radio::endReception() at Radio.cc:473 0x7fffd3fd6a72	
inet::physicallayer::Radio::handleReceptionTimer() at Radio.cc:255 0x7fffd3fd3e14

One way to deal with this issue is to check in Radio::{start,continue,end,abort}Reception if the transmitter radio is still registered at the RadioMedium. This solution has been proposed by Nico Brüttner.
What would be the preferred way by INET core developers to solve this problem?

levy:
Thanks for bringing this up, although I'm not sure how to properly fix this issue. If we were only considered with the numerical output of simulations, then I would say:

For a signal which already completely left the transmitter and which just propagates on the medium, I would say that the signal should not be asked to return the transmitter at all. All transmitter data that is relevant for any further processing should have been already copied to the signal.

For other signals which are being transmitted at the moment when the radio is deleted from the simulation, I would say that the signal propagation should be changed to propagate the transmission abort event. And again, all transmitter data that is relevant for further processing should have been already copied to the signal.

So, I guess what I'm saying is that the transmitter should not be checked on the radio medium, because it should not be used at all. At least that's the case for the evolution of the simulation.

On the other hand, for various reasons it might be desirable to keep a reference in the signal to the transmitter. For example, for the sake of logging or debugging, e.g. knowing where the signal originates from.

Currently, the ITransmission::getTransmitter() function's documentation says that it never returns a nullptr. If we allow removing *and* deleting radios at any moment, then it's clearly wrong.

Do you really need to delete those radios as opposed to just removing them from the medium?

riebl:
[...]

PR #252, opened 2017-05-22
Prevent dereferencing dangling Radio pointer in RadioMedium

riebl:
This pull request relates to #250.
I have discarded the previous idea to replace IRadio pointers by radio ids for two reasons:
1) Ids from IRadio::getId() cannot be turned into cModule easily because it is not inherited from cModule::getId(). A workaround would require look-ups at RadioMedium.
2) I am not sure if it is possible to create any reasonable return value in DimensionalAnalogModelBase::computeReceptionPower at all when there is no transmitter radio available any more.

Hence, I have picked up and modified the initial idea by Nico Brüttner. For faster lookups, std::set is used instead of std::vector to store IRadio* in RadioMedium. Fingerprints remain unchanged and wireless tests are still running fine.

### DIFF
--- a/src/inet/physicallayer/common/packetlevel/Radio.cc
+++ b/src/inet/physicallayer/common/packetlevel/Radio.cc
@@ -414,23 +414,27 @@ RadioFrame *Radio::createRadioFrame(cPacket *packet) const
 void Radio::startReception(cMessage *timer, IRadioSignal::SignalPart part)
 {
     auto radioFrame = static_cast<RadioFrame *>(timer->getControlInfo());
-    auto arrival = radioFrame->getArrival();
-    auto reception = radioFrame->getReception();
-// TODO: should be this, but it breaks fingerprints: if (receptionTimer == nullptr && isReceiverMode(radioMode) && arrival->getStartTime(part) == simTime()) {
-    if (isReceiverMode(radioMode) && arrival->getStartTime(part) == simTime()) {
-        auto transmission = radioFrame->getTransmission();
-        auto isReceptionAttempted = medium->isReceptionAttempted(this, transmission, part);
-        EV_INFO << "Reception started: " << (isReceptionAttempted ? "attempting" : "not attempting") << " " << (IRadioFrame *)radioFrame << " " << IRadioSignal::getSignalPartName(part) << " as " << reception << endl;
-        if (isReceptionAttempted)
-            receptionTimer = timer;
+    if (medium->hasRadio(radioFrame->getTransmission()->getTransmitter())) {
+        auto arrival = radioFrame->getArrival();
+        auto reception = radioFrame->getReception();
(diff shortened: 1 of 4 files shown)


levy:
I thought that we are going to change the reception process in a way that allows receiving signals despite the transmitter has been already removed from the medium while the signal was propagating. This change simply ignores a signal at the receiver if the transmitter has been already removed removed from the medium, which is clearly just a kludge. I understand that this change avoids the crash, but you could also avoid that by carefully removing a radio after it stopped transmitting for a certain amount of time. I also understand your (2) point, namely that it's not trivial to change the reception process in a way that completely avoids querying the transmitter radio. We would have to copy all relevant information to the signal during the transmission. I'm reluctant to include this change in it's current form due to the above reasoning.

riebl:
I am very well aware that this pull request is only a crude workaround. Are you willing to accept a patch which includes severer changes at ITransmission? From my point of view, only antenna properties are really missing in ITransmission right now: ITransmission::getTransmitter()::getAntenna() is a prerequisite for calculating antenna gains.
Extending the lifetime of IAntenna (and consequently IMobility) compared to the related IRadio seems to be unfavourable to me considering that AntennaBase is a cModule (for good reason) and thus can hardly be managed by a kind of smart pointer. Not even speaking about the IMobility trouble arising then…
Honestly, I am not sure how to solve this issue in a nice way. Maybe a std::shared_ptr<IAntennaProperties> returned by IAntenna::getProperties()? This shared_ptr could be included in ITransmission and be used for gain computations instead.

levy:
Hmm, this issue seems to be much deeper than we anticipated. The transmitter cannot foresee all potential receivers, especially those which may not yet be known by the medium, at the time of the transmission. So if we allow the transmitter to be deleted while the signal propagates, then we have to make sure that the transmitter antenna data is available by some other means at the receiver during the reception. This is required to compute directional selectivity for the transmitter antenna.

In general, any physical property of the transmitter that may affect the reception of a signal, and which may be altered over time (either by the destruction of the transmitter or simply by changing it on the transmitter) must be copied to the transmission. This includes position, orientation, modulation, encoding, power, frequency, etc. **and antenna data**. I think shared_ptr would only solve the destruction case but not the modification case. Besides, there's no shared_ptr right now in INET.

I think right now it seems to be easier for your case to delay the removing of the radio. You could delay the removing until a safe moment, when all of the radio's transmissions have been already removed from the medium.

As for the long run, I think this issue should be solved by copying all relevant data to the transmission so it stands on its own feet.

riebl:
I was thinking of std::shared_ptr (INET makes already use of C++11, so this one is available) to reduce the overhead caused by each Transmission creation. Most of the time the antenna properties are quite static. Maybe a custom smart pointer with copy-on-write semantics could help us here? Such a smart pointer would also behave correctly for the modification case you have mentioned.

levy:
Should I close this one too after the merge of #266, or is there any value left in it?

riebl:
I close this pull request because it is completely superseded by #266. Thank you for all the good comments!

PR #261, opened 2017-07-03
Fixed erroneous CSMA queue length check which resulted in one extra queue entry

i-tek:
Queue frames were inserted even the queue length had been reached already.
This error reduced the number of queue drops and increased the performance.

I don't know whether that extra queue entry was originally intended as an "extra processing buffer"-imitation but I guess it was not (nothing documented). Therefore I assume its an error.


koalo:
Can now be closed, was merged as f723568.

levy:
Thanks. The potential infinite loop in GPSR was also fixed in integration.

PR #262, opened 2017-08-08
Transmitter and Antenna snapshots in Transmission objects

riebl:
These changes eliminate the dangling pointer issue #250. I have been able to remove hundreds of radios simultaneously based on this pull request without segfault whereas master branch fails.

### DIFF
--- a/src/inet/physicallayer/antenna/ConstantGainAntenna.cc
+++ b/src/inet/physicallayer/antenna/ConstantGainAntenna.cc
@@ -44,6 +44,11 @@ std::ostream& ConstantGainAntenna::printToStream(std::ostream& stream, int level
     return AntennaBase::printToStream(stream, level);
 }
 
+std::shared_ptr<IAntennaSnapshot> ConstantGainAntenna::createSnapshot()
+{
+    return std::make_shared<Snapshot>(gain);
+}
+
 } // namespace physicallayer
 
 } // namespace inet
--- a/src/inet/physicallayer/antenna/ConstantGainAntenna.h
+++ b/src/inet/physicallayer/antenna/ConstantGainAntenna.h
@@ -31,13 +31,22 @@ class INET_API ConstantGainAntenna : public AntennaBase
 
   protected:
     virtual void initialize(int stage) override;
+    class Snapshot : public IAntennaSnapshot
+    {
+    public:
+      Snapshot(double gain_) : gain(gain_) {}
+      virtual double computeGain(const EulerAngles direction) const override { return gain; }
+
+    protected:
+      double gain;
+    };
 
   public:
     ConstantGainAntenna();
 
     virtual std::ostream& printToStream(std::ostream& stream, int level) const override;
     virtual double getMaxGain() const override { return gain; }
-    virtual double computeGain(const EulerAngles direction) const override { return gain; }
+    virtual std::shared_ptr<IAntennaSnapshot> createSnapshot() override;
 };
 
 } // namespace physicallayer
--- a/src/inet/physicallayer/antenna/CosineAntenna.cc
+++ b/src/inet/physicallayer/antenna/CosineAntenna.cc
@@ -48,7 +48,17 @@ std::ostream& CosineAntenna::printToStream(std::ostream& stream, int level) cons
     return AntennaBase::printToStream(stream, level);
 }
 
-double CosineAntenna::computeGain(const EulerAngles direction) const
+std::shared_ptr<IAntennaSnapshot> CosineAntenna::createSnapshot()
+{
+    return std::make_shared<Snapshot>(maxGain, beamWidth);
+}
+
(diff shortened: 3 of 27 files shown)


levy:
I'm on holiday right now but I'll take a look at this when I'm back.

Thank you!

riebl:
Did you have time to have a look on this patch by now?

levy:
Ah, sorry for the long delay. Actually, I looked through this patch and it has some merits. Nevertheless, I haven't made up my mind because I have some issues with it. I don't really like the idea of TransmitterSnapshot class and also having shared_ptr in the interfaces. I guess it would be ok to have shared_ptr in the implementations though.

Hmm, could you please look at my proposed solution (based on yours):
 - add ITransmission::getRadioMedium()
 - add ITransmission::getTransmitterId()
 - replace implementation of ITransmission::getTransmitter() with a lookup using the medium and the id, it may return nullptr if not found
 - do something similar with IReception
 - refactor all getTransmitter() usage to expect nullptr too, use getRadioMedium() directly where possible, use getTransmitterId() where it's sufficient
 - refactor IAntenna to provide mobility and a separate directional selectivity?
 - refactor all antenna modules to use a separate class (which can be simply duplicated) that implements the directional selectivity interface (computeGain)
 - copy directional selectivity to both transmission and reception when they are created

This allows the transmission and reception models to be completely free of any changes in the transmitter and receiver radios including deletion. I'm aware of the fact that this solution doesn't use shared pointers but we may add that later, we will see. Processing a signal already allocates a number of objects at each receiver so I'm not sure if allocating one more (as opposed to using shared pointers) is such a big problem. Even more so, because I think separating the analog signal representation from the transmission class (similarly to how its done in LayeredTransmission) is desirable to avoid having so many different transmission classes in the future.

Let me know what do you think about my proposed solution.

riebl:
I have already been afraid you might dislike the std::shared_ptr approach, though I don't quite understand why this should be of any concern even in interfaces.
Without any profiling it is hard to foresee the impact of copying the antenna gain parameters for each transmission. I am not worrying about the additional memory consumption, however, it might execute slower due to worse cache coherence compared to shared_ptr. This is speculative as said before, though.

Nevertheless, adapting your proposal should not be too complicated. In the analog models, lines such as "const IAntenna *transmitterAntenna = transmitterRadio->getAntenna();" would then become "const IAntennaGain *tranmitterAntennaGain = transmission->getTransmitterAntennaGain();" assuming IAntennaGain is the "directional selectivity" interface.
I will rework my patch accordingly as soon as time permits. :-)

levy:
I think your suggestion with IAntennaGain sounds right. I'm not saying that using shared_ptr here is wrong. Actually we are using shared_ptr a lot in the new packet API of INET 4.0. I guess I just would like to make it one step at a time to better see the costs & benefits. 

If we are at it, we don't really plan to release a new version of the INET 3.x branch, because we are not actively working on it anymore. I guess, you don't want to migrate to the upcoming INET 4.0 version just to be able to have this change. Hmm, maybe we should end up having a 3.6.1 only with this change?

riebl:
Maybe you can have a look at https://github.com/riebl/inet/commits/antenna_gain
There are a few things I don't like that much but where I would like to hear your opinion:

- IAntennaGain requires a duplicate() method. Currently, it returns a raw pointer but transfers ownership. I would prefer to use std::unique_ptr<IAntennaGain> for this purpose.
- If use of std::unique_ptr is okay in INET, antenna implementations might store their AntennaGain using this smart pointer and only create these objects during initialize(). This would eliminate the need for default constructors using NaN values.
- transmission->getTransmitter()->getMedium() chains might be replaced by transmission->getMedium() [not yet implemented, but IRadioMedium* is now included in each transmission]. Follow-up commit?

I still believe that IAntenna::getGain() should simply return std::shared_ptr<const IAntennaGain> and eliminate the aforementioned duplicate() method. Copies would only be required if any directional selectivity parameters changed during runtime, which is currently not the case for any antenna implementation.

Currently, my plan is to wait for INET 4.0 to stabilise and migrate Artery later on. However, we remain using 3.6 for some already on-going studies. If you don't want to release a 3.6.2 version I can maintain this change locally.

levy:
I looked through your branch and it looks good to me. I agree with all of your points including eliminating the duplicate() method using shared_ptr instead. Replacing transmission->getTransmitter()->getMedium() chains in a follow-up commit would be nice. One change which I didn't like is the AntennaParameters of InterpolatingAntenna. The AntennaGain subclass is already a friend of InterpolatingAntenna, so there's no need to have a separate AntennaParameters class. You could simply move all utility functions into the AntennaGain as protected, use the default constructor and fill in the data later.

Let me know what do you think. In general, I'm going include these changes (send me a pull request please) in the INET integration branch. This is the branch that will be turned into INET 4.0 before the end of this year.

levy:
I close this pull request because #266 is better and it has been merged.

PR #264, opened 2017-08-11
Edcaf: fix aifs calculation for customised aifsn

riebl:
Bug: Edcaf always takes the fallback value from getAifsNumber(ac) instead of the aifsn parameter when given.


levy:
Thanks for the fix, merged to master.

PR #266, opened 2017-09-15
IAntennaGain integration

riebl:
Fixing dangling transmitter radio pointer of #250. This pull requests represents a revised version #262 targeting the integration branch.


levy:
This pull request looks good. It would be nice to have a regression test which fails now but passes when this pull request is merged. I merge it anyway, thank you for your contribution!

PR #272, opened 2017-11-03
Minor fixes on various modules

MaximilianKoestler:
This fixes some smaller issues and adds a minor improvement to the IdealEpEnergyStorage.
The changes should be self-explanatory.

### DIFF
--- a/src/inet/networklayer/generic/GenericRoutingTable.cc
+++ b/src/inet/networklayer/generic/GenericRoutingTable.cc
@@ -145,13 +145,15 @@ void GenericRoutingTable::configureRouterId()
     if (routerId.isUnspecified()) {    // not yet configured
         const char *routerIdStr = par("routerId").stringValue();
         if (!strcmp(routerIdStr, "auto")) {    // non-"auto" cases already handled in earlier stage
-            // choose highest interface address as routerId
+            // choose lowest interface address as routerId
             for (int i = 0; i < ift->getNumInterfaces(); ++i) {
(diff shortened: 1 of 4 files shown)


levy:
I don't see the inconsistency between GenericRoutingTable and L3AddressResolver, the latter doesn't even support routerId for generic addresses. Other than that the pull request could be merged but I wait until this gets resolved either by removing it from the pull request or by providing a better justification.

src/inet/networklayer/generic/GenericRoutingTable.cc
```diff
@@ -145,13 +145,15 @@ void GenericRoutingTable::configureRouterId()
...
-            // choose highest interface address as routerId
+            // choose lowest interface address as routerId
             for (int i = 0; i < ift->getNumInterfaces(); ++i) {
                 InterfaceEntry *ie = ift->getInterface(i);
                 if (!ie->isLoopback()) {
                     L3Address interfaceAddr = ie->getGenericNetworkProtocolData()->getAddress();
-                    if (routerId.isUnspecified() || routerId < interfaceAddr)
+                    if (routerId.isUnspecified() || routerId < interfaceAddr) {
                         routerId = interfaceAddr;
+                        break;
```
ZoltanBojthe: I think, this "break" is not a good idea here.
Function returns first non-loopback address instead of the lowest address.

koalo:
This pull request is now split up in separate pull requests. The routerId commit is postponed. Thus, this pull request can be closed.

levy:
Thanks!

PR #275, opened 2017-11-24
consider sensitivity for computeIsReceptionPossible in APSKScalarReceiver

tobithiel:
The `APSKScalarReceiver` is using the `NarrowbandReceiverBase` to determine whether a reception is possible. This is only checking whether the carrier frequency and bandwidth are equal. The `FlatReceiverBase` has a more advanced check whether a reception is possible by checking against the `sensitivity` parameter of the receiver.
I changed the call to the parent module and added a test that checks whether the `sensitivity` is considered when deciding if a reception is gonna be attempted.
Let me know what you think!


levy:
You are absolutely right! This bug has been already fixed in integration. I will merge this into master but I think the test is unnecessary. This bug is really minor and there's no reason to fear from regression. Tests must be to maintained, they are not free. Anyway, thank you!

tobithiel:
@levy Ah didn't notice that it was already fixed in integration. Thank you :smile:

PR #276, opened 2017-12-04
tcp: serializer: Fix checksum computation

noackrob:
RFC 793, page 16, depicts that

> If a segment contains an odd number of header and text octets to be checksummed, the last octet is padded on the right with zeros to form a 16 bit word for checksum purposes.

The former implementation did not add the additional octed, which causes the computed checksum to be wrong.

Signed-off-by: Robert Noack <robert.noack@u-blox.com>

### DIFF
--- a/src/inet/common/serializer/TCPIPchecksum.cc
+++ b/src/inet/common/serializer/TCPIPchecksum.cc
@@ -34,8 +34,10 @@ uint16_t TCPIPchecksum::_checksum(const void *_addr, unsigned int count)
         count -= 2;
     }
 
-    if (count)
-        sum += *(const uint8_t *)addr;
+    if (count) {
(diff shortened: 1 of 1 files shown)


torokati44:
Thank you, nice catch!
However, this is already fixed in the integration branch:
https://github.com/inet-framework/inet/commit/7a78d15b9e4e63aa207ca594b4fea7e4d814ac32
I'm picking this onto the master branch instead of yours, just to avoid a merge-conflict later.

noackrob:
Sorry, I should have checked the integration branch. The other fix works fine for me, thank you very much.

torokati44:
No problem. We know that it is unusual to have the most recent state in, and perform most active work on, a branch other than `master`. This might change in the future, so this kind of confusion and duplicated work can be avoided more easily. Thank you once again.

PR #281, opened 2017-12-19
Added implementation of the X-MAC linklayer protocol

ApolloLV:
This is the implementation of Joaquim Oller for the MiXiM framework that I ported to INET as a part of my bachelor thesis.
It is useful in the context of wireless sensor network simulations when considering low power listening protocols.

### DIFF
--- a/src/inet/linklayer/xmac/XMacFrame.msg
+++ b/src/inet/linklayer/xmac/XMacFrame.msg
@@ -1,5 +1,5 @@
 //
-// Copyright (C) 2013 OpenSim Ltd
+// Copyright (C) 2017 Jan Peter Drees
 //
 // This program is free software; you can redistribute it and/or
 // modify it under the terms of the GNU Lesser General Public License
@@ -14,8 +14,6 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with this program; if not, see <http://www.gnu.org/licenses/>.
 //
-// author: Zoltan Bojthe
-//
 
 cplusplus {{
 #include "inet/common/INETDefs.h"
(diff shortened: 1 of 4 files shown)


levy:
Very nice! These are the kind of contributions we are looking for. We are going to look into this pull request and I hope merge soon. Do you think that you can help us out and also port it to INET 4.0 if requested? BTW, what does X stands for?

ApolloLV:
X does not really stand for anything. It is just another protocol from the MAC alphabet soup.
(www.st.ewi.tudelft.nl/~koen/MACsoup/)

levy:
Thank you anyway!

levy:
We (my colleague Gyorgy Szaszko and me) are creating a showcase (a short tutorial) about wireless sensor network MAC protocols, namely B-MAC, L-MAC, and X-MAC. The documentation for the showcase is still a work in progress, you can find the current version here:

https://inet.omnetpp.org/inet-showcases/wireless/sensornetwork/

Could you please help us with some questions?

1. Can XMac work with an asymmetrical slotDuration (i.e. the slotDuration value is not the same for all nodes in the network)?

2. Currently, we use XMac with ApskScalarRadio, which always adds a PHY preamble and PHY header to the signal. We are unsure about how the MAC preamble is supposed to be represented at the PHY?

- PHY preamble + PHY header + some MAC data representing the MAC preamble (optionally zero bits?). This is what we have right now but this is probably wrong.
- PHY preamble (some predefined waveform), no PHY header, no MAC data and no PHY overhead at all? Is this the right interpretation?

ApolloLV:
1. XMac could work, but my current implementation does not support asymmetry.
You could change [this line](https://github.com/inet-framework/inet/blob/4b415fead79e25a441dfda9cd5a6037d1bb82d9d/src/inet/linklayer/xmac/XMacLayer.cc#L324) so it schedules the message at the longest slot duration configured in the network instead of the slot duration of the current node.

PR #282, opened 2017-12-21
Fix retransmission bugs

ruengeler:
- The transmission counter was not advanced when the data was timer based retransmitted.
- When fast recovery was active and the cumulative TSN advanced, the miss indications were not increased.
- In PacketDrill the counting of Gap Reports was fixed.
- A PacketDrill test to test fast Retransmissions was added.


levy:
Thank you!

PR #286, opened 2018-01-26
Ieee802154: Set rxSetupTime

koalo:
The rxSetupTime for the IEEE 802.15.4 MAC was still set to 0s.
This is now set according to the datasheet of the ATmega256RFR2.

Signed-off-by: Florian Kauer <florian.kauer@koalo.de>

### DIFF
--- a/src/inet/linklayer/ieee802154/AT86RF231.ned
+++ b/src/inet/linklayer/ieee802154/AT86RF231.ned
@@ -1,0 +1,28 @@
+//
+// Copyright (C) 2014 Florian Meier
+// Copyright (C) 2013 OpenSim Ltd.
+//
+// This program is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public License
+// as published by the Free Software Foundation; either version 2
+// of the License, or (at your option) any later version.
+//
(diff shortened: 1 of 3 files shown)


levy:
Is this chip the de-facto standard or just one implementation among many?

koalo:
It is used at lot and the transceiver equivalent to the AT86RF231 that is used in even more devices. Of course, it is not the only available transceiver, but the value will not differentiate a lot for the others.

levy:
On the long run, I think, it would be better to have an AT86RF231.ned which configures every parameter and submodule according to the chip datasheet and remove the chip specific default values from the generic module which should only contain what is mandated by the standard.

koalo:
Yes, that might be a clean option. But what to use as default, then?

levy:
I think if there's no sensible default value in the standard then it should be left empty. This indicates for the user that she have to think about it. I'm not saying that this how things are done in INET in general, because unfortunately it's not. But I'm not sure that we are going in the right direction with this patch.

koalo:
The exact value should not be that important, just that it is larger than 0s. Anyway, I added a AT86RF231.ned with that value.

PR #296, opened 2018-02-28
bugfix: corrected memory management issue in radio module

goinsj2010:
Corrected a fatal bug in the radio module that can occur if the separateReceptionParts mode is used.

This fixes an invalid delete(void*) attempt that occurs if a packet reception ends while the radio is still switching from RX --> another mode. Previously the code was trying to use abortReception(cMessage*) on a packet that had already been processed with endReception(cMessage*). If your simulation was unlucky enough to encounter this case the simulator would segfault.


levy:
Indeed, thank you for the good explanation.

PR #300, opened 2018-03-15
Integrate sctp

ruengeler:
Port the SCTP deserializer
Port the EXT interface
Test the emulation examples extclient and traceroute
Run all SCTP tests


levy:
Thank you Irene for the pull request!

I admit that I tried to rebase and merge this pull request but I have some issues with it. It would be nice to separate SCTP related changes from other changes. We accept all SCTP related changes without deep review but changes to other INET parts must be reviewed first. Mixing the two set of changes makes it hard to review and understand the motivation behind the patches. I also noticed quite a few print statements left in, some EV statments replaced with std::cout, etc. these are considered bad practice in general. It would be nice to either have proper EV statements in place or remove them altogether.

Could you please fix these issues and send another pull request?
Thank you very much!

PR #315, opened 2018-04-24
802.11 VHT mode

aarizaq:
Implementation of VHT mode. There are some things that should be checked like the physical header length. I have used the same definitions present in HT mode. 

The computation of the error header, I don't understant this
int chunkLength = bitLength - b(htMode->getHeaderMode()->getHTLengthLength()).get();
I have used
int chunkLength = bitLength

I don’t understand what mean qbpsk, I have replaced by bpsk.

The band module should support a field with the bandwidth used. For HT and VHT and future standard is important. I have avoid the modification of the class including “5GHz & 20MHz”, “5GHz & 40MHz” … definitions, but it isn’t elegant.

Due to that it is possible to have several configuration with the same bit rate, the selected is the first defined.
The bit rate in the selection is rounded to Mb/s with a decimal (like in the standard tables)

I have included an example

And sorry, I have merge the branch with the lates modification in inetgration and github sends the modifications.


levy:
Ah, thank you very much! This is a great contribution, especially the error model part, because that was missing a long time ago.

Instead of merging your pull request, I've split it and committed several patches in your name. I also did a few small modifications such as renaming the bands, fixing includes. I've still left out 2 parts which I'm not sure about:

1. QBPS (from the standard): quadrature binary phase shift keying (QBPSK): A binary phase shift keying modulation in which the binary data is mapped onto the imaginary (Q) axis.

2. Ieee80211TransmitterBase: we need a better way to select the mode based on the band, bandwidth and bitrate.

aarizaq:
There is a problem in the error model if you use QbpskModulation,

In
 double Ieee80211NistErrorModel::getOFDMAndERPOFDMChunkSuccessRate(const ApskModulationBase *subcarrierModulation, const ConvolutionalCode *convolutionalCode, unsigned int bitLength, double snr) const

and

double Ieee80211YansErrorModel::getOFDMAndERPOFDMChunkSuccessRate(const ApskModulationBase* subcarrierModulation, const ConvolutionalCode* convolutionalCode, unsigned int bitLength, bps grossBitrate, Hz bandwidth, double snr) const

It is necessary to change 
    if (subcarrierModulation == &BpskModulation::singleton) {
To 
    if (subcarrierModulation == &BpskModulation::singleton || subcarrierModulation == &QbpskModulation::singleton) {

It is necessary to fix the computation of Ieee80211VhtPreambleMode::duration 
        virtual const inline simtime_t getDoubleGIDuration() const { return 2 * getGIDuration(); } // GI2
        virtual const inline simtime_t getLSIGDuration() const { return getSymbolInterval(); } // L-SIG
        virtual const inline simtime_t getNonHTShortTrainingSequenceDuration() const { return 10 * getDFTPeriod() / 4;  } // L-STF
        virtual const inline simtime_t getHTGreenfieldShortTrainingFieldDuration() const { return 10 * getDFTPeriod() / 4; } // HT-GF-STF
        virtual const inline simtime_t getNonHTLongTrainingFieldDuration() const { return 2 * getDFTPeriod() + getDoubleGIDuration(); } // L-LTF
        virtual const inline simtime_t getNonHTSignalField() const { return 4E-6; } // L-SIG
        virtual const inline simtime_t getVHTSignalFieldA() const { return 8E-6; } // VHT-SIG-A
        virtual const inline simtime_t getVHTShortTrainingFieldDuration() const { return 4E-6; } // VHT-STF
        virtual const inline simtime_t getVHTSignalFieldB() const { return 4E-6; } // VHT-SIG-A

const simtime_t Ieee80211VhtPreambleMode::getDuration() const
{
    // 21.3.4 Mathematical description of signals
    simtime_t sumOfHTLTFs = getSecondAndSubsequentHTLongTrainingFielDuration() * (numberOfHTLongTrainings);
    return getNonHTShortTrainingSequenceDuration() + getNonHTLongTrainingFieldDuration() + getLSIGDuration() + getVHTSignalFieldA() + getVHTShortTrainingFieldDuration() + sumOfHTLTFs + getVHTSignalFieldB();
}


There is a problem if you use dynamic node creation with the destination address. This is a concept proof to solve the address with dynamic node creation. This can be adapted to any source type, and can recompute the list of destination address when a  node is created/deleted.

[base.zip](https://github.com/inet-framework/inet/files/1954851/base.zip)

levy:
I added QBPSK as you suggested, but I'm not quite sure about the rest of the changes in VHT. Could you please send a pull request instead?

I added a TODO regarding the dynamic node creation and address resolution issue.

PR #316, opened 2018-04-27
Use CONGESTION instead of RETRY_LIMIT_REACHED for backoff limit reached

koalo:
The retry limit is for a different purpose than the backoff limit for CSMA/CA in IEEE 802.15.4 so they should be tracked separately. I think CONGESTION matches the case of a channel that was sensed busy several times very good. An alternative would be to add a BACKOFF_LIMIT_REACHED.


levy:
Thanks!

PR #317, opened 2018-04-27
Make it possible to extend GPSR

koalo:
I would like to be able to make some changes in the way GPSR chooses the next hop, thus one suggestion is to make some methods virtual and protected and create a derived class. Another alternative would be to integrate it directly in GPSR and add a switch. What do you think about it?


levy:
It depends on what are those changes. Making methods virtual is not a problem.

koalo:
Please have a look at this:
https://github.com/openDSME/inet-dsme/tree/master/src/routing/gpsr

levy:
Could you tell me more about the differences? What's the goal? I don't mind to make those methods virtual by the way.

koalo:
Would be nice, but unfortunately, I currently have no time to assemble a new patch.
By the way, Tobias just uploaded his thesis https://doi.org/10.15480/882.1664 (Chapter 4).

PR #318, opened 2018-04-27
Computation of VHT physical header

aarizaq:


levy:
Thank you!

I've squased and rebased the differences in your branch into intergration. I left out the run script, because that's not needed anymore. You should run '. setenv' in the INET folder and examples can be run using the scripts under the 'bin' folder.

I hope all is included now, so I close this pull request.

PR #319, opened 2018-04-27
Only use formatString and parseFormat if we build with visualizer

koalo:


levy:
Thanks for the bug report and the patch!

I decided to move the StringFormat class under common to remove the dependency instead.

koalo:
That is fine, thanks!

PR #322, opened 2018-05-03
Reordering

ManiAm:
small refactoring

re-ordering lines in creating routes in AODV and RIP to reflect the order that routing entry fields are printed


levy:
I cherry-picked your refactoring patch. Thanks!

PR #325, opened 2018-05-04
Example: Video streaming with TCP, UDP and the DASH technique.

andersonandrei:
It's an example of video streaming with TCP, UDP protocols and the DASH techinique.
To do the video streaming with TCP we developed the TCPBasicVideoStream and with the DASH we developed the TCPVideoStreamCliApp and both are in src/inet/applications/tcpapp. 

I already discussed this pull request in the Fórum of the project :
https://groups.google.com/forum/#!topic/inetframework-contrib/oQEFU5vt0jQ.

There, Levy recommended me to do this way.

Best regards,
Anderson Andrei :)


levy:
Thanks for taking the time to send a pull request!

Unfortunately it seems that there was a slight misundertanding :( because you submitted the pull request against the INET master branch, which is the old INET 3.x version. If you reread my original suggestion, I explicitly asked for the integration branch, which is the upcoming INET 4.0 version.

If you don't want to port your pull request to the INET 4.0 version, then we will do it eventually, but certainly not before the INET 4.0 version is released.

I'm sorry about the misunderstanding.

andersonandrei:
Thanks for answer!

Sorry about the misunderstanding. I already did a new pull request, this time to the integration branch.

Best regards,
Anderson Andrei :)

PR #327, opened 2018-05-07
Direct routes

ManiAm:
adding the new 'addDirectRoutes' parameter to Ipv4NetworkConfigurator


levy:
Thank you for the pull request!

I cherry-picked your patch to avoid also having the merge patches. I slightly modified it to be more consistent with the rest of the class.

PR #329, opened 2018-05-09
Example (to the INET 4.x): Video streaming with TCP, UDP and the DASH technique.

andersonandrei:
It's an example of video streaming with TCP, UDP protocols and the DASH techinique.
To do the video streaming with TCP we developed the TCPBasicVideoStream and with the DASH we developed the TCPVideoStreamCliApp and both are in src/inet/applications/tcpapp.

I already discussed this pull request in the Fórum of the project :
https://groups.google.com/forum/#!topic/inetframework-contrib/oQEFU5vt0jQ.

There, Levy recommended me to do this way.

Best regards,
Anderson Andrei :)


levy:
Thanks for the pull request!

Did you actually tried to compile and run your contribution using this INET version? Because it seems to me that it would not compile, or even if it would compile, it would not run properly.

There are many changes in the new INET 4.0 version compared to the old INET 3.x versions. That's why I was asking for a pull request against the new version, because otherwise we would have to do the heavy lifting.

andersonandrei:
Hello !

I am trying to update my example and modules to the INET 4.0 and it is compiling and running for me now. The UDP example is working, but I have some problems with the "Chunk settings" in my TCP simulations:

The simulation compile and run, but stop in some seconds with the message: 
     " Returning an incomplete chunk is not allowed according to the flags : 1 ..."
I can not correct it and I have no idea how to do it. Can you help me, please ?

Best regards,
Anderson Andrei :)

levy:
Ah, sorry for the late answer, I was busy with other things and I forgot this thread.

Incomplete chunk means that you are trying to peek/pop more data from a packet than what is available. Normally this shouldn't happen, but it all depends on what you are trying to achieve. Without more context, it's hard to say what is the cause.

I also noticed that the two apps are very similar, it would be nice to have one instead and some parameters to get the behavior of both. Maybe it would be easier if you could just make this pull request up-to-date, then we could merge it into integration and we would come back to fixing it.

Note: don't update your INET to the latest integration now, because it requires a not yet released omnet.

andersonandrei:
Hello !

To have one instead and some parameters to get the behavior of both look me a goo idea, I could learn how to do it later.

I am confused about the pull request. Your first answer in the pull request was : 

	“There are many changes in the new INET 4.0 version compared to the old INET 3.x versions. That's why I was asking for a pull request against the new version, because otherwise we would have to do the heavy lifting.”
	
	So I downloaded the latest development version (3.99.3 for OMNeT++ 5.3) at the INET page and update my first pull request project to this version. 
	What is the correct version of INET should I download ?
	After to download and adapt, should I pull request my files separated? For example:
	- A pull request for TcpBasicVideoStream.* (TCP)	
	- A pull request for TcpVideoStream.* (DASH)
	- The scenario example

Sorry about my misunderstanding.

Best Regards,
Anderson Andrei :)

ZoltanBojthe:
Hi!
I tryed adding this pull request for inet integration. My topic branch in inet is the topic/andersonandrei-bz.
I fixed the compiler errors in the code for current integration, but the TcpVideoStreamServerApp missed for tcp-based simulations.

andersonandrei:
Hello !

I corrected the TCP and DASH server "typename". But the same error persists.

@ZoltanBojthe, thank you so much for creating the branch :).
How could I update the omnetpp.ini of my simulation there?

Best Regards,
Anderson Andrei.

FireChemist:
Will this be compatible with OMNeT++ 5.1 and INET 3.5?  I'm developing a video streaming application over TCP running on a LTE network (hence using Simulte).

levy:
No, we are not working on INET 3.x branch anymore. This could land in INET 4.x eventually, but the branch which contains the related changes has not been touched since a while.

FireChemist:
@andersonandrei 
git pull
git checkout topic/andersonandrei-bz

Git pull will get the new branches and git checkout will switch to the branch that he created.

PR #346, opened 2018-06-07
Improve the performance of the computeIntersection function of the Prism shape

panagis:
This function copies the points instead of referencing them.
In this fix, I replace the copies with references.

Using this fix in a large scale system of CAVs simulation,
the simulation time dropped from **357** minutes to **218** minutes.


levy:
Thanks! I merged your patch into the integration branch.

PR #348, opened 2018-06-11
fix uint snd_cwnd in NewReno having an overflow wrap around becoming …

freiheitsnetz:
…very large


levy:
I assume that this is the same pull request as #349, so I close this.

PR #349, opened 2018-06-11
fix uint snd_cwnd in NewReno having an overflow wrap around becoming …

freiheitsnetz:
…very large


levy:
This patch seems to be right, but I don't know enough about the topic to be sure. Could you give a little bit of more reasoning?

Thank you!

PR #350, opened 2018-06-11
introduce statistics for 802.11 mac pending queue

freiheitsnetz:
Would be nice to have statistics for the pending queue of the 802.11 mac layer


levy:
I think emitting another packet dropped signal is superfluous and too expensive. I think the current drop signal contains enough information to decide if the packet was dropped from the pending queue (source == Dcf && reason == QUEUE_OVERFLOW). As for the queue length signal, I think it would be better to have a packetEnqueued or something similar to how other queues notify subscribers. In the future we might actually want to replace the internal queue with an explicit queue module.

PR #357, opened 2018-07-10
ieee80211: support EtherType protocol discrimination in 5.9 GHz band

riebl:
According to E.2.3/E.2.4 of IEEE 802.11-2016, stations in the 5.9 GHz band have to use "EtherType Protocol Discrimination" (EPD) instead of IEEE 802.2 (LPD) commonly used for other bands. Unfortunately, there is no packet field to differentiate them so a `LlcProtocolTag` is added to each packet for dissecting both variants seamlessly. If this tag is missing and there is no information available about the employed band, it falls back to IEEE 802.2.

### DIFF
--- a/examples/wireless/nic/omnetpp.ini
+++ b/examples/wireless/nic/omnetpp.ini
@@ -225,3 +225,17 @@ description = "IEEE 80211 mac with scalar radio"
 extends = AbstractIeee80211Mac, AbstractIeee80211DimensionalRadio
 description = "IEEE 802.11 mac with IEEE 802.11 dimensional radio"
 
+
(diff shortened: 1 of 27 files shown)


levy:
Thank you for your pull request! This seems to be a non-trivial change, it will take some time to review but at a first glance it seems to be ok. I plan this to be included in the next release.

riebl:
@levy I am happy to hear this. Let me know when you need any assistance.

levy:
Thanks for this pull request! I have merged this into master. I'm going to add a few cosmetic changes such as renaming Ieee80211Llc to IIeee80211Llc, because this is a module interface.

PR #360, opened 2018-07-16
Adding RIP passive mode

ManiAm:
In most of the RIP implementations, the local interfaces of a router need to manually be added. If a local interface is not added to RIP, then RIP completely ignores the network attached to that interface. In other words, RIP will not send any updates from that interface AND will not include that network in the RIP updates sent from other interfaces.

Many implementations, provide an option to set an interface into _passive_ mode. User can set a local interface as passive to prevent the transmission of routing updates through that interface BUT still allows that network to be advertised in routing updates that are sent out to other interfaces.

Now coming back to the INET implementation. Usage of NoRIP is confusing, because it acts like both scenarios above. In this pull request I separated PASSIVE from NoRIP.

**NoRIP:** means the RIP completely ignores that local interface.  Not only it does not send any updates from that interface, but also does not advertise the network attached to that interface from other interfaces.

**Passive:** means the RIP does not sent any updates from that interface but it still advertises the network attached to that interface from other interfaces.

Edit: I sent the PR to the integration branch. I do not know if future changes are applied into the master directly or integration.


levy:
I pushed your contribution into the master branch using git rebase. I had to split up your patch to separate the smallest possible set of changes which actually change the fingerprints of RIP examples.

Could you please verify that the last one (which changes fingerprints) is definitely correct?

Thank you very much!

ManiAm:
Yes, everything looks good.

levy:
As it turns out the patch broke a few fingerprint and module tests, because the network interfaces in routers towards hosts were not configured for RIP by default. So we had to add something like the following line to the examples (pushed into master):

`<interface source="R?" dest="H?" mode="PASSIVE" metric="1"/>`

Do you think that it is acceptable to require users to configure such network interfaces as above? I'm asking to be sure that this change is the right thing to do.

In the future, you may want to run fingerprint tests for RIP in the folder 'tests/fingerprint' to check if your changes affect them:
./fingerprinttest --debug -m rip

You can use the following to get some help:
./fingerprinttest -h

ManiAm:
If a user does not provide any RIPconfig file, then all interfaces of a router are set into 'split horizon with poisoned reverse'. This means that non of the interfaces are in passive mode and RIP sends updates to all the interfaces. However, if the user wants to set one or more interfaces into passive, then he needs to provide a new RIPConfig and set one or more interfaces into passive similar to commit 400ffa7. This is the same approach taken by Cisco and Juniper.

Look at the following example RIP configuration in a Cisco router.



The first line is analogous to useRip = true :)
The second and third lines are analogous to the default RIPConfig that adds all active interfaces of a router into RIP.
The last line specifically instructs the RIP to put an interface into passive.

>> Do you think that it is acceptable to require users to configure such network interfaces as above?

Yes. If a user want a passive interface, then he needs to set it manually.

>> In the future, you may want to run fingerprint tests for RIP in the folder 'tests/fingerprint' to check if your changes affect them

Ok, Thanks!

levy:
Thanks for the explanation, this makes sense.

PR #363, opened 2018-07-19
Ospf improvements

ManiAm:
Kudos to the person who implemented OSPF which I believe is one of the most complex protocols in INET.

[commit ed94d05]

Adding `startupTime` parameter to OSPF similar to RIP to control the start up time of the OSPF process in routers. Currently all OSPF processes start at the begining of simulation which is not realistic. I set the default value to 0s though, but ideally it should be a random value between like 0s and 5s or such.

[commit 825606a]

There is no need to add directly-connected routes into the routing table by the OSPF since they are already added before even OSPF is initialized. A routing protocol is responsible for learning remote routes. Similar to RIP or any other routing protocols, OSPF is now adding remote networks only.

[commit d500781]

If `useAdminDist` parameter is set to `true` in a routing table (IPv4 or IPv6) then the administrator distance is shown along with metric for each route (for both RIP and OSPF routes)

2018-07-sdsd-41-02

[commit 83dcdd8]

RIP and OSPF now set the admin dist of remote networks that they have learnt. A very interesting scenario is now running both RIP and OSPF on routers in an internetwork. Routing tables of routers would have two routes to the same destination, but one is learnt from OSPF and one is learnt from RIP. The routers prefer the route learnt from OSPF over RIP, because it has a lower admin dist and is appeared upper in the table. I have simulated this scenario in Step7 of this example and it's very cool!

https://github.com/ManiAm/inet/blob/INETex/examples_INETex/static_dynamic_routing/omnetpp.ini

Check the routing table of router0 for instance and note that remote routes learned by OSPF are always placed before the routes to the same network learned by RIP. Now that admin dists are also shown, the user know what is happening.

[commit 613c114]

The OSPF standard does not explicitly specify any metric so the admin/user can set any metric he wants. However in many implementations (including Cisco and Juniper), OSPF cost calculation is based on a reference bandwidth of 10e8 as following:

cost of a link = 10e8 / bandwidth of link (bps)

I have added this feature so that if `interfaceOutputCost` is set to 0 then OSPF calculates the cost of each interface based on the formula above.

[commit 45afcf6]

Moving a couple of functions around to better reflect the execution flow and improve code readability.

[commit 83f119c and 2a9d50e]

Rename a few class names and types to reduce ambiguity.

Interface --> OspfInterface
InterfaceState --> OspfInterfaceState
InterfaceStateType --> OspfInterfaceStateType
InterfaceEventType --> OspfInterfaceEventType
RoutingTableEntry --> OspfRoutingTableEntry

[commit 968439c]

Using auto to improve code readability.

[commit 4d98c4e]

Moving the OSPF-related information to the end of each route. The OSPF routing table is now more consistent with other routing tables including IPv4 and RIP. Also empty fields like next hop are denoted by * rather than `<unspecified>`.

[commit 89d1994]

Printing the link-state database (LSDB) before SPF tree construction. This DB shows the global topology in an AS or area that the router has learnt so far. This is a valuable output for the user to understand how the SPF tree is constructed.

[commit b72270b and ca05a92]

Important data structures used by OSPF process are added into WATCH. This free the user from digging into the source code to find out the latest values of internal OSPF data structures.

[commit 1992ac9 and 6951455]

fix issue #367
6951455: forgot to logic AND with network mask

[commit 7bf3f39, 1f29f4f]

adding list of OSPF neighbors + neighborhood status for each. This is super useful in figuring out OSPF adjacency establishment especially in DR/BDR selection.

[commit 0e669b7 and 636d7d1]

Adding protocol-specific information to the main IPv4 routing table similar to other routing protocols.

[commit 707e66f]

a multi-access interface performs the DR/BDR election process after the Wait timer is expired, but no log is generated on who was selected. This commits print the result of the election.

[commit cd2fe7b]

small refactoring: re-arranging and grouping related OSPF params

[commit 78d7545]

Adding an area section in the OSPF config file is mandatory. However in many scenarios, the user wants to simulate a flat topology where all OSPF routers belong to the backbone (area 0.0.0.0). In this case the OSPF config file can be simplified by not adding any area id for interfaces as well as area definition attribute. In this commit, if no area is specified in the OSPF config, then the default area 0.0.0.0 will be used. This is the first step in simplifying the OSPF config.

[commit 61cc338, d74eff4 and 1970466]

I have added OSPF interface mode similar to the one I added for RIP:
Active: The interface is active (default mode)
Passive: The interface is advertised but no OSPF message is sent out from this interface
NoOSPF: The interface is not advertised and is completely ignored by the OSPF.

[commit 47ba6b0 and a30f783]

So far the user had to specify an OSPF config for each scenario. A lazy user like me wants to rely on the default settings and setup an OSPF scenario by just setting the useOspf parameter to true (similar to RIP that a default RIPconfig is used when no RIPconfig is provided). This commit adds this capability. The OSPF automatically detects the network that an interface is attached to (broadcast, point-to-point) and sets its type. In addition to the default config, OSPF configuration such as below is also possible:

```
<OSPFASConfig>

  <Router name="R1" RFC1583Compatible="true">
    <BroadcastInterface ifName="eth0" interfaceOutputCost='454' />
    <BroadcastInterface ifName='eth[*]' areaID='0.0.0.0' interfaceOutputCost='0' /> \
  </Router>
  
  <Router name="**" RFC1583Compatible="true">
     <BroadcastInterface ifName='eth[*]' areaID='0.0.0.0' interfaceOutputCost='0' search='relax' /> \
     <PointToPointInterface ifName='ppp[*]' areaID='0.0.0.0' interfaceOutputCost='0' search='relax' /> \
  </Router>

</OSPFASConfig>
```
It instructs OSPF to configure R1 differently than other routers. Moreover, on R1, it sets parameters for eth0 differently than other Ethernet interfaces. Note that eth0 must exists on router R1 otherwise OSPF throws an error. One can relax this behavior by setting the search param to `relax`. 

Also note that if you do not specify an interface in the Router section, then that interface will not be added into the OSPF. This is identical to adding the interface under the Router section and setting its `interfaceMode` to NoOSPF.

[commit adba6b8, c04f2ba and 6184601]

Adding the checksum field to the OSPF messages (that was missing before) + adding OSPF serializer class. I have also tested the serializer by capturing PCAP files and opening it with wireshark. Calculated checksum by wireshark is marked as correct and no malformed packet is produced.

[commit f4db61c]

removing the annoying new line that causes unnecessary new lines:

[commit 7ecdfdf, cbba9ab]

once the OSPF routing table is rebuilt, all existing OSPF routes are removed and then updated routes are added. Here is a sample output log:

The user had to go over all the changes to find out if any new route has been added or any existing routes is deleted. This commit makes the life much easier for users and prints a summary of route changes (if any):

[commit 02bbde7, 0a44a13, 187cf82, 0d80f86, 0eb582e, d1a8f84, 4d4fb2a, 27fd158, a8faee2, a27e9c0, c07dbcf]

small improvements

[commit 7ceb5f5, ba8421e, f78d5cc, 7a7767a, cbb1c4c, fff421a]

OSPF's reaction to link breakage was slow. In fact RIP converges faster than current OSPF implementation. I noticed that OSPF is not subscribed to interface create/delete/change signals. This means that when a link breaks, OSPF waits for the 'router dead interval' (20s by default) and announces its neighbor dead. These commits fix this problem. As soon as a directly-connected link breaks, OSPF reacts and rebuilds the routing table.

[commit ddb78e3]

fixing a major segmentation fault. In summary, when a link is broken, OSPF rebuilds the SPF tree, but forgets to clear some of the pointers in the old SPF tree. One of these pointers is `parent` relation. The parent of a node should be set to null if that node is not present in the newly constructed SPF tree. If not set to null, then segmentation fault might occur in some scenarios with link breakage.

[commit 22dece3]

When the DR goes down, BDR is supposed to take over without the need to re-election. Apparently, in INET when DR goes down, BDR detects that DR is dead (after not receiving any Hello for router dead interval), but does nothing! This commit fixes this issue.

[commit 9343d3a]

If the user does not specify any output cost for an OSPF interface, then the cost is calculated based on the bandwidth.

[commit 5549e37]

small improvement

[commit c6999ea, 68f5082]

Change `status = "Advertised"` to `advertise="true"` or `advertise="false"`. Also this attribute is not mandatory any more. If the user does not specify this attribute, then the address is advertised by default. In almost all cases, we want to advertise the address. Also I update the OSPF examples (Note that in all OSPF examples, even full test, addresses are advertised).

[commit bc2e328, 350faff. c85d051, f8ed4c0]

improvements

[commit ee325a3]

Summary LSAs were not being advertised correctly to non-backbone areas after topology change.

[commit 28294d6]

Routing table is updated correctly after receiving an AS-External LSA.

[commit f1db5cd]

**No!** A router connected to multiple areas is not necessarily a ABR. At least one of its interfaces should be connected to the backbone.
**No!** non-ABR routers are not allowed to generate summary LSAs!
**No!** The current implementation populates the routing tables with wrong routes in scenarios where a router connects to multiple areas, but not the backbone! such as `Step17` in [here](https://github.com/ManiAm/inet/blob/INETex/examples_INETex/routing_protocol_ospf/omnetpp.ini). 

The following `if`  is so wrong in so many levels ...

https://github.com/inet-framework/inet/blob/f62a4cead396a2d083645b3e02824fcf4f4a5d8f/src/inet/routing/ospfv2/OspfConfigReader.cc#L462

https://www.youtube.com/watch?v=31g0YE61PLQ

[commit 0653915]

Non-backbone areas can now be configured as stub. A stub attribute was already defined in the Ospf.ned module, but there was no implementation.

https://github.com/inet-framework/inet/blob/4568110adacaaeb9f85a21967adce87a08855768/src/inet/routing/ospfv2/Ospf.ned#L37

On top of that, some of the Ospfv2 examples, such as below, show scenarios with OSPF stub, but interestingly the Stub attribute is not even used in the config. Even if used, it has no effect. So I was thinking what exactly they are showing! Any ways, here is an example that shows how to set a non-backbone area as stub:

https://github.com/ManiAm/inet/blob/49fbcef32f3a6543828204a8940bf872b8234309/examples_INETex/routing_protocol_ospf/omnetpp.ini#L390

There is another form of OSPF stub, defined by OSPF RFC 3101, called `Not-So-Stubby Area (NSSA)` and two more are defined by Cisco (`Totally Stubby` and `Totally NSSA`). These types are also interesting, but they are not in my high-priority feature list.

-------------------------------------------

All examples are working as expected:

-------------------------------------------


levy:
Thanks for you great efforts! I'm following your changes but I haven't review them yet. I'll come back when I do.

ZoltanBojthe:
Hi! I created the [ManiAm-ospf_improvements](https://github.com/inet-framework/inet/commits/ManiAm-ospf_improvements) branch in inet-framework/inet and try your changes.
(I rebased your pull request to current master and added some fixes for follow changes in master and fixing the ospf lifecycle and some other trivial problems)

Some BGP+OSPF examples does not work correctly.
For example, see the `/examples/bgpv4/BgpUpdate -c config1`: the A.bgp module doesn't receive any BgpUpdate packet.
In my last [cbf4d44](https://github.com/inet-framework/inet/commit/cbf4d44a3855ebf5781a0f86d7ae927e8d242ef2) commit, you can see the all examples where fingerprint changed due to your modifications.

ManiAm:
>> Some BGP+OSPF examples does not work correctly.

Thanks! I will take a look at them. I have more suggestions/commits on OSPF and I am going to create a new PL from now on based on your new commits.

ManiAm:
[commit 9343d3a]

If the user does not specify any output cost for an OSPF interface, then the cost is calculated based on the bandwidth.

[commit c6999ea, 68f5082]

Change status = "Advertised" to advertise="true" or advertise="false". Also this attribute is not mandatory any more. If the user does not specify this attribute, then the address is advertised by default. In almost all cases, we want to advertise the address. Also I update the OSPF examples (Note that in all OSPF examples, even full test, addresses are advertised).

[ToDo]

1. When BDR goes down, reelection is not done! 
2. Working on #372 
3. Working on OSPF tutorial step-by-step examples similar to RIP:

ManiAm:
[commit bc2e328, 350faff. c85d051, f8ed4c0]

improvements

[commit ee325a3]

Summary LSAs were not being advertised correctly to non-backbone areas after topology change.

[commit 28294d6]

Routing table is updated correctly after received an AS-External LSA.

[ToDo]

1. When BDR goes down, reelection is not done! 
2. Working on #372 
3. Working on OSPF tutorial step-by-step examples similar to RIP:

ManiAm:
>> We don't understand the 44fd82992045aa96f7dc0928aa4660ab0c866fe6 patch. We looked at #367 and we understand the described problem. But why would OSPF deal with "routes" which lead to a local interface at all? Isn't OSPF supposed to create routes only for non-local destinations? This looks strange, so maybe we need another solution. Unfortunately, several patches depend on this one, so we're kind of stucked at this point.

The OSPF's behavior is a little weird (especially when comparing with other routing protocols) but it is not wrong! OSPF supports many network/link types and that's where most of the confusion and complexity starts. For example in the network I mentioned in #367, the link between router0 and router1 is a point to point (P2P) link. OSPF adds the two ends of the P2P link with /32 mask into the routing table namely 10.0.0.17 and 10.0.0.18 and not the network (10.0.0.16) as in other protocols like RIP, EIGRP, etc. If you change the link type to broadcast, then OSPF adds the 10.0.0.16 network into the routing table like other routing protocols with the expected metric/cost.

The reason that P2P links are treated like above by OSPF is due to the way the SPF tree is constructed. The SPF tree of the same above network is like this:

ddfdf

OSPF uses the Diskstra to find the shortest path to all square nodes (N1, N2 and N3 are broadcast networks whereas N4 to N9 are networks formed by the end of each P2P link). Now what the shortest path from R1 (router 0) to N4 (10.0.17/32)? Here is the shortest path:

R1 --> R5 --> R4 --> N2 --> R2 --> N4 : total cost: 1+1+1+0+10=13

That's why you see the first route (as highlighted in the picture) that says: network 10.0.0.17/32 is reachable with cost of 13. So OSPF is behaving correctly and can find all the shortest path for us. All these paths are stored in the internal OSPF routing table, but we should not allow OSPF to add that route into the IPv4 routing table. Why? Because, 10.0.0.17 is a directly-connected route and router0 can reach it with cost of 1 which is less that 13.

levy:
We tried to continue the merging of this pull request into the master branch, but we actually failed. The main reason is that we discovered certain regressions using fingerprint testing, and we lack the required knowledge to fix the affected protocols.

To give you an example, the 825606a7916656731dbe7391a30495e0f5eac845 patch is fairly trivial to understand, but it breaks some of the BGP examples and one PIM IPTV example too. We looked into the details and found that the BGP protocol works differently due to the above patch. We clearly don't have the required knowledge to actually fix these issues right now. We would have to dig deep into OSPF and BGP to understand what should happen compared to what's really going on.

Well, you have already put a tremendous effort into improving RIP, OSPF, and BGP. So, what I'm asking may be too much, but we think you actually have the required knowledge to fix these issues. You certainly have more knowledge regarding OSPF and BGP than we have. We don't want to push more work on you, but if we are going to do it, then it will take a lot more time. We might even need to postpone this work for the sake of other priorities.

Anyway, to help you with fingerprint testing in general, I quickly wrote a new [chapter](https://inet.omnetpp.org/docs/developers-guide/) in the INET Developer's Guide. If you are willing to use fingerprint testing to avoid regression and sort of prove correctness, then this documentation can be very hopeful. I know it's an extra burden, but this would make the merging your changes so much easier.

I also added 'ospf' tags to several examples in the CSV files, so you can easily and quickly run only the required subset of fingerprint tests while developing OSPF.

You can set up the testing environment with (I guess you already know this):
cd ~/workspace/omnetpp
. setenv
cd ~/workspace/inet
. setenv
cd tests/fingerprint

You can run all tests regarding OSPF with (all 23 of them should pass):
./fingerprinttest -m ospf

Meanwhile, we are going to pick your patches individually, if they are acceptable changes and they don't introduce regressions.

Let me know, if I can help in any way.

ManiAm:
Sure. I will try to check them all, however, I am also working full-time and my free time fluctuates significantly. What I can say for sure is that many of the OSPF scenarios were giving me wrong results. I often cross-check the output with Cisco packet tracer by creating the same exact scenario there and verify the results. My OSPF PL corrects many of the existing issues and all the scenarios below give me a correct/reasonable output that match the output in Cisco packet tracer. 

Anyways, I am looking at the BGP code these days (on my free time) and I will try to verify the code + output with the Cisco packet tracer.

>> Meanwhile, we are going to pick your patches individually ...

Ok. I believe the majority portion of the PL could be safely applied without breaking the existing fingerprints.

Let's make INET great again! :)

ManiAm:
[commit 0653915]

Non-backbone areas can now be configured as stub. A stub attribute was already defined in the Ospf.ned module, but there was no implementation.

https://github.com/inet-framework/inet/blob/4568110adacaaeb9f85a21967adce87a08855768/src/inet/routing/ospfv2/Ospf.ned#L37

On top of that, some of the Ospfv2 examples, such as below, show scenarios with OSPF stub, but interestingly the Stub attribute is not even used in the config. Even if used, it has no effect. So I was thinking what exactly they are showing! Any ways, here is an example that shows how to set a non-backbone area as stub:

https://github.com/ManiAm/inet/blob/49fbcef32f3a6543828204a8940bf872b8234309/examples_INETex/routing_protocol_ospf/omnetpp.ini#L390

There is another form of OSPF stub, defined by OSPF RFC 3101, called `Not-So-Stubby Area (NSSA)` and two more are defined by Cisco (`Totally Stubby` and `Totally NSSA`). These types are also interesting, but they are not in my high-priority feature list.

levy:
Ok, this is starting to get really funny with the Not-So-Stubby Area. :-)

levy:
I have merged another bunch of patches into master from this PL. There are still quite a few which are left out.

levy:
This PR has been manually merged into INET master, it will be released with INET 4.1.
If you have additional changes to OSPF, please open a new PR.
Thank you for your valuable contribution!

PR #374, opened 2018-08-19
ipv4config: reserving all host ids inside a network

ManiAm:
**Issue 1:** 

Consider the following network:

```
network sampleNetwork
{
    @display("bgb=1915.2001,741.60004");

    types:
        channel PppLink100M extends DatarateChannel
        {
            delay = 5us;
            datarate = 100Mbps;
        }

        channel PppLink10M extends DatarateChannel
        {
            delay = 5us;
            datarate = 10Mbps;
        }

    submodules:
        configurator: Ipv4NetworkConfigurator {
            @display("p=160.35374,123.55125");
        }
        visualizer: IntegratedCanvasVisualizer {
            @display("p=425.85748,123.55125");
        }
        scenarioManager: ScenarioManager {
            @display("p=712.39124,123.55125");
        }
        R2: Router {
            @display("p=959.4937,573.0675");
        }
        switch3: EtherSwitch {
            @display("p=160.35374,570.4387");
        }
        R4: Router {
            @display("p=410.085,573.0675");
        }
        R3: Router {
            @display("p=709.7625,573.0675");
        }

        switch2: EtherSwitch {
            @display("p=1264.4287,570.4387");
        }
        switch5: EtherSwitch {
            @display("p=709.7625,378.54");
        }
        R1: Router {
            @display("p=1519.4175,573.0675");
        }
        switch4: EtherSwitch {
            @display("p=959.4937,378.54");
        }
        switch1: EtherSwitch {
            @display("p=1782.2925,567.81");
        }
        R5: Router {
            @display("p=1264.1412,379.0025");
        }
    connections:
        R2.pppg++ <--> PppLink100M <--> R3.pppg++;
        R3.pppg++ <--> PppLink100M <--> R4.pppg++;
        R4.ethg++ <--> Eth100M <--> switch3.ethg++;
        R3.ethg++ <--> Eth100M <--> switch5.ethg++;
        R2.ethg++ <--> Eth100M <--> switch2.ethg++;
        R1.ethg++ <--> Eth100M <--> switch2.ethg++;
        switch4.ethg++ <--> Eth100M <--> R2.ethg++;
        switch1.ethg++ <--> Eth100M <--> R1.ethg++;
        switch2.ethg++ <--> Eth100M <--> R5.ethg++;
}
```
With following configuration:

```
[Config Step9]
description = "sample example"
network = sampleNetwork

*.configurator.addStaticRoutes = false

# Enable OSPF on all routers
*.R*.hasOspf = true

# Visualizer settings
*.visualizer.interfaceTableVisualizer.displayInterfaceTables = true
```
The static IPv4 assigned to subnets are overlapping:

The LAN formed by switch2 is assigned network id of 10.0.0.8/29 that covers IP addresses of 10.0.0.9 to 10.0.0.15 (broadcast). This range overlaps with the network 10.0.0.12/30 assigned to interface eth1 of router R2 that covers 10.0.0.13 to 10.0.0.15 (broadcast). This is not wrong, but can be problematic in some scenarios (please check motivation section). 

If I add a host into the 10.0.0.8/29 network, then eth1 interface of R2 is assigned address of 10.0.0.17/30. 

The user might need to reserve all host ids inside a network even though there are not enough interfaces inside that network. For example if host id is 3 bits, then we can assign IPv4 addresses to maximum of 2^3-2=6. Now if there is only 3 interfaces on that network, we assign IPv4 address to those three interfaces, but I want to reserve the remaining three addresses. In other words, I want to instruct the IPv4configurator to assume that those three addresses are also taken and it should not use them in other neighboring networks.

**Motivation:** 

I noticed that OSPF does not add a newly-discovered remote route into the routing table because the new route overlaps with an existing route in the routing table that has IPv4 overlap. Because of these lines:

https://github.com/inet-framework/inet/blob/a817ccda8db4c89c14f1a3d04b9c36c8926cd831/src/inet/routing/ospfv2/router/OspfArea.cc#L1746

More specifically, the newly-discovered route is 10.0.0.12/30, but there exist route 10.0.0.8/29 in the routing table. Since 10.0.0.12/30 overlaps with the existing route, OSPF will not add it, which is wrong!

This PL adds the feature to the ipv4 configurator using the `useStrictDisjunctSubnetAddressParameter ` parameter. For example, consider the following example, consisting of a subnet of 4 nodes. Subnet mask is /29 thus 6 IPv4 addresses can be assigned in total. 4 of them are assigned to the existing interfaces and since `useStrictDisjunctSubnetAddressParameter` is true, the two remaining IPv4 addresses are reserved.

Here are a few simulation scenarios to test the new parameter:

https://github.com/ManiAm/inet/tree/INETex/examples_INETex/ipv4configurator

**Issue 2:** 

the interface eth1 of router R2 should be assigned an IPv4 address of the form x.x.x.x/32 because that is the only interface on that network. I guess the ipv4configurator thinks that switch4 is also part of the network and hence uses /30 as network mask, but switch is transparent to IP.

### DIFF
--- a/src/inet/networklayer/configurator/ipv4/Ipv4NetworkConfigurator.cc
+++ b/src/inet/networklayer/configurator/ipv4/Ipv4NetworkConfigurator.cc
@@ -530,6 +530,7 @@ void Ipv4NetworkConfigurator::assignAddresses(Topology& topology)
                     if(itt == assignedAddressToInterfaceEntryMap.end()) {
                         assignedAddressToInterfaceEntryMap[completeAddress] = nullptr;
                         assignedInterfaceAddresses.push_back(completeAddress);
+                        EV_DEBUG << "reserving Ipv4 address " << Ipv4Address(completeAddress) << ".\n";
                     }
                 }
             }
--- a/src/inet/networklayer/configurator/ipv4/Ipv4NetworkConfigurator.cc
+++ b/src/inet/networklayer/configurator/ipv4/Ipv4NetworkConfigurator.cc
@@ -68,6 +68,7 @@ void Ipv4NetworkConfigurator::initialize(int stage)
     if (stage == INITSTAGE_LOCAL) {
         assignAddressesParameter = par("assignAddresses");
         assignDisjunctSubnetAddressesParameter = par("assignDisjunctSubnetAddresses");
+        useStrictDisjunctSubnetAddressParameter = par("useStrictDisjunctSubnetAddress");
         addStaticRoutesParameter = par("addStaticRoutes");
         addSubnetRoutesParameter = par("addSubnetRoutes");
         addDefaultRoutesParameter = par("addDefaultRoutes");
@@ -520,6 +521,17 @@ void Ipv4NetworkConfigurator::assignAddresses(Topology& topology)
 
                 // remove configured interface
                 unconfiguredInterfaces.erase(find(unconfiguredInterfaces, compatibleInterface));
+            }
+
+            if(assignDisjunctSubnetAddressesParameter && useStrictDisjunctSubnetAddressParameter) {
+                for(int n = 1; n < (1 << (32-netmaskLength))-1; n++) {
+                    uint32 completeAddress = networkAddress + n;
+                    auto itt = assignedAddressToInterfaceEntryMap.find(completeAddress);
+                    if(itt == assignedAddressToInterfaceEntryMap.end()) {
+                        assignedAddressToInterfaceEntryMap[completeAddress] = nullptr;
+                        assignedInterfaceAddresses.push_back(completeAddress);
+                    }
+                }
             }
 
             // register the network address and netmask as being used
--- a/src/inet/networklayer/configurator/ipv4/Ipv4NetworkConfigurator.h
+++ b/src/inet/networklayer/configurator/ipv4/Ipv4NetworkConfigurator.h
@@ -124,6 +124,7 @@ class INET_API Ipv4NetworkConfigurator : public NetworkConfiguratorBase
     // parameters
     bool assignAddressesParameter = false;
     bool assignDisjunctSubnetAddressesParameter = false;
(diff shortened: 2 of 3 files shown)


avarga:
Hi Mani, thank you for recent contributions, we very much appreciate it! Levente is on holiday this week, we'll review and merge your changes when he's back.

levy:
Issue 1: I understand the motivation and agree with it. I'm not sure if we really need two parameters though. It seems to me that the original motivation behind the assignDisjunctSubnetAddresses was the one you are describing. Otherwise I don't see how this parameter supposed to work. My simple guess is that the configurator was not prepared for subnet addresses which are prefixes of each other.

Issue 2: ppp0 of R1 and R2 are on the same link, so there are two addresses in that subnet, plus the all 0s and all 1s, which is 4. So the netmask is /30. The configurator doesn't know about PPP, so it can't figure out that it would just as be good with /32 too.

So my suggestion is that we should have one parameter instead, and it should be the one which was already there. Do you agress?

ManiAm:
>> So my suggestion is that we should have one parameter instead, and it should be the one which was already there. Do you agree?

That would be even better. Frankly, when I first encountered this issue I was thinking why the `assignDisjunctSubnetAddresses` does not do this in the first place.

levy:
After reviewing your changes and understanding the original configurator code (it took some time :-), I came up with a different solution. 

I didn't introduce a new parameter, because we have already agreed on it. I rather fixed the old parameter, which I think was supposed to work the way you think it should have. Your solution seems to be less optimal to me in the sense that it may require adding all potential network interfaces to the list of already assigned network interfaces. The list may grow very large depending on the specified bits of the netmask (e.g. /16).

The https://github.com/inet-framework/inet/tree/topic/ipv4networkconfigurator branch contains the solution. I verified it on the example you brought up and it fixes the issue. I also tested this branch and it passes all fingerprint tests.

If you could also verify and agree that this branch really solves the issue you raise, then I will merge this into master.

levy:
The change has been merged into master.

PR #375, opened 2018-08-20
Ipv4config: isolated networks can be configured separately

ManiAm:
IPv4 configurator builds a topology from the network and uses that topology to assign IPv4 addresses to each interface. Imagine the following scenario that consists of two isolated networks: These two networks are completely detached from each other, but the topology class detect it as a single big network. This means that the IPv4 addresses assigned to the second isolated network (on the bottom) will be the continuation of the first network (on the top).

In some scenarios, the user wants to treat these two networks isolated and expects the IPv4 addresses to be completely independent from each other. This commit adds a new parameter to the IP configurator called `configureEachGroupSeparatly`. If it is set to true, then the IPv4 addresses are assigned like the following:

The topology class can now detect the number of network groups and pass this information to the IPv4 configurator so that it treats each group independently. 

You can find the above example in `Step9` of the following:

If the network is like the following:

Then it says:

In this case setting `configureEachGroupSeparatly` to true does not have any effects because there is only one group in the scenario. Also note that this approach works in a mixed wired-wireless scenarios and it only considers the groups in wired portion of the network.

**Update:** well, you can use a better name than `group` that one might misinterpret for IP multicast group! `NumDisjointNetworks` can be another suggestion.

### DIFF
--- a/src/inet/common/Topology.cc
+++ b/src/inet/common/Topology.cc
@@ -197,12 +197,15 @@ void Topology::extractFromNetwork(bool (*predicate)(cModule *, void *), void *da
 {
     clear();
 
+    int groupCounter = 0;
+
     // Loop through all modules and find those that satisfy the criteria
     for (int modId = 0; modId <= getSimulation()->getLastComponentId(); modId++)
     {
         cModule *module = getSimulation()->getModule(modId);
         if (module && predicate(module, data)) {
             Node *node = createNode(module);
+            node->setGroupId(++groupCounter);
             nodes.push_back(node);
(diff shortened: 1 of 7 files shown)


levy:
The quality of your descriptions and patches amaze me. Thank you again!

I would like to change the group name to something else in a separate patch as you suggested.

ManiAm:
Thanks! When a software package like INET is very well-designed, people get motivated to contribute. I have been using OMNET++/INET for 6 years now, but recently decided to contribute back :)

PR #377, opened 2018-08-25
visualizer: fix seg fault. Make sure the ptr is not null

ManiAm:
A simple pointer check prevents SEG fault.
To reproduce the error, try to run the the following scenario:

https://github.com/inet-framework/inet/tree/master/examples/ospfv2/areas

And turn on the interface visualization.

`*.visualizer.interfaceTableVisualizer.displayInterfaceTables = true`

### DIFF
--- a/src/inet/visualizer/linklayer/InterfaceTableCanvasVisualizer.cc
+++ b/src/inet/visualizer/linklayer/InterfaceTableCanvasVisualizer.cc
@@ -120,7 +120,7 @@ void InterfaceTableCanvasVisualizer::addInterfaceVisualization(const InterfaceVi
         displayString.setTagArg("t", 0, getVisualizationText(getInterfaceEntry(interfaceVisualization)).c_str());
         displayString.setTagArg("t", 1, "l");
     }
-    else
+    else if(interfaceCanvasVisualization->networkNodeVisualization)
         interfaceCanvasVisualization->networkNodeVisualization->addAnnotation(interfaceCanvasVisualization->figure, interfaceCanvasVisualization->figure->getBounds().getSize(), placementHint, placementPriority);
 }
 
@@ -131,7 +131,7 @@ void InterfaceTableCanvasVisualizer::removeInterfaceVisualization(const Interfac
     auto gate = displayWiredInterfacesAtConnections ? getOutputGate(interfaceVisualization) : nullptr;
     if (gate != nullptr)
         gate->getDisplayString().setTagArg("t", 0, "");
-    else
+    else if(interfaceCanvasVisualization->networkNodeVisualization)
         interfaceCanvasVisualization->networkNodeVisualization->removeAnnotation(interfaceCanvasVisualization->figure);
 }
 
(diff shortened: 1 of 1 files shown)


levy:
This should be fixed now in master. Actually, you have to add to visualizers, one for each subnet, and it should work out of the box.

ManiAm:
>> Actually, you have to add to visualizers

In the old code, adding only one visualizer was enough. I added only one visualizer in the network module of the following example:

https://github.com/inet-framework/inet/tree/master/examples/ospfv2/areas

I am able to see the visualization for each of the subnets when I double-click on them.

levy:
I don't know how could it actually happen, but even if it did work for some reason, this was clear unintended. Maybe this worked because for InterfaceTableCanvasVisualizer, because it used gate display strings instead of figures. This is basically a kludge, and it would not work with all the other visualizers.

All visualizers have a visualizationTargetModule (formerly called visualizerTargetModule) which determines the canvas where they display information. This means that one visualizer can only display information on one canvas.

Visualizers also make possible, by the way, to have multiple different visualizations of the same network via using multiple differently configured visualizers and having different visualizationTargetModules. This can be useful, for example, to have a logical (i.e. compact) and physical (i.e. real coordinates) visualization of the network.

ManiAm:
**Update:** I understand that adding the visualizer into the top level only is not the right way to go as you discussed, however if a user does that he/she gets SEG FAULT. Checking for null pointers is necessary I guess (if avoiding SEG FAULT is a must). After git pull the latest changes today and re-run the same old scenario (with visualizer at the top level), my program crashes.

levy:
Indeed, the example still crashes if I put the visualizer at the top level. Of course, any combination of NED files, and any parameterization using INI files should not lead to a crash. Nevertheless, I think it is generally wrong to silently ignore the user request, in this case the request to display certain interfaces from the interface table. So how about putting guards 813b3f747d37b515d37f79126293b02024974ddb there which stops the simulation with a proper error message instead?

ManiAm:
Yes, I agree. That's the correct way to go. Thanks!

ManiAm:
Interface table visualization causes problem for one of my scenarios (again) and I do not know how to fix it. I am not sure if it is a bug or a miss-configuration! Please take a look at OSPF examples in here:

Step18 throws a visualization error:

levy:
The problem is caused by the default setting of the InterfaceTableVisualizerBase module. The nodeFilter is set to '*' by default which matches the full name of all nodes no matter how deep they are. So basically, the top level InterfaceTableVisualizer tries to visalize all interfaces from all nodes down the module tree. This obviously fails because there are no visual representations of the nodes in N1 and N2 subnets.

Try adding the following line:
*.visualizer.interfaceTableVisualizer.nodeFilter = "not(fullPath(*.N*))"

I admit that this is not optimal, but I don't have a really good idea how to solve it right now. Unfortunately, it can't be solved by filtering for node names. Full path filtering doesn't help either, because it does a substring match right now.

ManiAm:
Thanks for the hint, but it still does not work. I copied your line into the configuration, but I am still getting the same error.

PR #381, opened 2018-09-09
Bgp improvements

ManiAm:
[commit 85b294b]

small refactoring to make the BGP code consistent with other routing protocols

[commit a7332e8, 879a997]

BGP.cc was a monolithic class and is now split into the following smaller classes. This is persistent with what the OSPF code is already organized.

Bgp
BgpConfigReader
BgpRouter

[commit b946548, 04e6793]

I went over the example in `bgpv4/Bgp2RoutersInAS` and noticed that the output is not correct. 

The first problem: four different external routes are being advertised by the OSPF ASBR, and one of them
is missing. Other routers do not have any routing entry pointing to that external route. Commit b946548 fixes this problem.

The second problem: when multiple BGP messages are sent to the TCP for transportation, some of them 
were surprisingly get disappeared on the receiver side. Apparently, the BGP messages get encapsulated correctly into a TCP segment (picture below), but due to a bug in the BGP receiver, only the first chunk is extracted. 

Commit 04e6793 fixes this problem.

And yes, these two modifications change the fingerprint!

Similarly, example `Bgp3Routers` produces wrong output. Many of the external routes are not getting advertised. The above two commits fix the issue.

[commit dab7d39]

Multiple NED parameters are introduced in the BGP module in order to control the behavior of the BGP protocol. The user can change these parameters to turn of/off different aspects of the BGP. You can find multiple examples here:

https://github.com/ManiAm/inet/blob/INETex/examples_INETex/routing_protocol_bgp/omnetpp.ini

The last scenario is interesting. It shows a mini-Internet consisting of three ASes. The left-most and right-most ASes (Stub ASes) are running RIP and the middle AS (transit AS) is running OPSF internally. Communication between ASes are done with BGP. At time t=100 a host in one AS pings another host connected to the other end.

-------------------------------------------

All examples are working as expected:

https://github.com/ManiAm/inet/blob/INETex/examples_INETex/routing_protocol_bgp/omnetpp.ini

-------------------------------------------


ManiAm:
I went over the example in `bgpv4/Bgp2RoutersInAS` and noticed that the output is not correct. 

The first problem: four different external routes are being advertised by the OSPF ASBR, and one of them
is missing. Other routers do not have any routing entry pointing to that external route. Commit b946548 fixes this problem.

The second problem: when multiple BGP messages are sent to the TCP for transportation, some of them 
were surprisingly get disappeared on the receiver side. Apparently, the BGP messages get encapsulated correctly into a TCP segment (picture below), but due to a bug in the BGP receiver, only the first chunk is extracted. 



Commit 04e6793 fixes this problem.

And yes, these two modifications change the fingerprint!

Similarly, example `Bgp3Routers` produces wrong output. Many of the external routes are not getting advertised. The above two commits fix the issue.

ManiAm:
[ToDo]

A cool simulation scenario is to build a small "Internet" consisting of three Autonomous Systems (AS). Each AS uses its own routing protocol internally. For example AS1 uses RIP, AS2 uses OSPF and AS3 uses EIGRP. All these three ASes are connected with BGP. Inside AS1 user X is connected to an IEEE 802.11 access point and ping a server on the other side of the small Internet in AS3. Complicated scenarios could be built to measure the performance of TCP congestion control, QoS measurement in the presence of differential or integrated services. This could be the one single scenario that demonstrates the power of INET in simulating the (wired) Internet.

levy:
Thank you again for the quality work! I have merged most of this pull request into master. I only left out the two patches which actually change the fingerprints. I'll return to this pull request later. Meanwhile, you may want to rebase your pull to make it look simpler.

levy:
In b946548, contrary to what the comment says, you also negate the condition. I don't really understand this patch, why did it work before and why does it work after applying it? I'm sorry if I sound nitpicking.

ManiAm:
>> The 04e6793 patch is quite suspicious to me, because you are not supposed to process chunks with checking what getChunkType returns. In fact, calling peekAll is probably wrong, because it could simply return a BytesChunk in which case you are not going to be able to process. In other words, processing must be done based on the actual data of the packet instead of the current representation. I feel like there's a missing piece of information there.

Using peekAll method and iterate over chunks was the only solution came to my mind.

ManiAm:
>> In b946548, contrary to what the comment says ...

levy:
This PR has been manually merged into INET master, it will be released with INET 4.1.
If you have additional changes to BGP, please open a new PR.
Thank you for your valuable contribution!

PR #382, opened 2018-09-13
improve implementation of TwoRayInterference path loss

riebl:
I have revisited my initial port (from Veins to INET) of the TwoRayInterference path loss model. It can handle now vertical as well as horizontal antenna polarization. Furthermore, I believe it is far more readable now due to restructured code and a lot more comments.


levy:
Thanks!

PR #383, opened 2018-09-13
Topic/inet4 for ansa

screw:
I hope I did not f-ed up when I merged master into topic/inet4-for-ansa during the relayUnit transition.

The L2 protocol above relayUnit needs to register MAC address that will be compared to frame's destination address to be passed UP.
To make it easier for me, STP registerAddress is hard-coded in relayUnit initialize method and CDP and LLDP (in ANSA) use a direct call to relayUnit -> **PLEASE** suggest a better solution for STP and I will copy it with CDP and LLDP in ANSA.

Also, (R)STP now needs `hasStp = true` to be activated. It does not longer depend on `spannigTreeProtocol != ""`.

Most fingerprint tests passed, but because of the `hasStp = true` change in the test setup, the Ieee8021d failed. (At least I assume that's the reason.)

Also included is Mac specific library name in bin/inet.


screw:
I just noticed the duplicate registration when multiple interfaces with EtherEncap are present eg any router.

ZoltanBojthe:
We are thinking on this problem. It's not trivial.

screw <notifications@github.com> ezt írta (időpont: 2018. szept. 14., P,
11:44):

ZoltanBojthe:
I rebased the topic/inet4-for-ansa to master and created a new topic branch with your commits and some fixes on the top of it: screw-topic/inet4-for-ansa. fingerprints correct on topic/inet4-for-ansa, but stp/rstp fingerprints failed/errored on screw-topic/inet4-for-ansa. I try to fix it at Monday

ZoltanBojthe:
Merged to topic/inet4-for-ansa

PR #384, opened 2018-09-14
Screw topic/inet4 for ansa

screw:
I fixed (from ERROR to FAIL) at least some of the RSTP specific tests from /examples/ieee8021d/ but `STP-LinkReconnect`still ERRORs.


ZoltanBojthe:
Merged, STP also fixed in topic/inet4-for-ansa

PR #398, opened 2018-11-25
Specifying join/leave time to/from a multicast group

ManiAm:
The Ipv4NetworkConfigurator is not only responsible for assigning IPv4 addresses to network interfaces, but also assigns multicast address. However, the current implementation only supports joining to a multicast address at the beginning of the simulation. 

In general, a node can join a multicast address at time t1 and leave the same multicast address at time t2. This PL adds two new parameters to the Ipv4NetworkConfigurator module. `joinTime` and `leaveTime`. The user can use these two parameters to specify when a node should join or leave a group. 

[Config Step5] shows an example:

https://github.com/ManiAm/inet/blob/INETex/examples_INETex/multicasting/omnetpp.ini

```
*.configurator.config = xml("<config> \
			<interface among='host* R0' address='10.x.x.x' netmask='255.x.x.x'/> \
			<interface hosts='**' address='x.x.x.x' netmask='255.x.x.x'/> \
  			<multicast-group hosts='host0' address='224.0.1.1' joinTime='5' leaveTime='20' /> \
  			<multicast-group hosts='host2' address='224.0.1.2' joinTime='12' leaveTime='25' /> \
       </config>")
```

Note that existing INET examples that use the multicast-group will not get affected. Because if joinTime is not specified, then 0s is assumed (node joins the group at the beginning of the simulation) and if leaveTime is not specified, then infinity is assumed (node never leaves the multicast group).

**Edit:** It is obvious that the leaveTime should always be after the joinTime. In the above configuration, one can set leaveTime before the joinTime. We get the following cryptic error:

Calling groupLeave on an interface when that interface has not joined to that group, should show a more meaningful error. In general, it is not the job of the Ipv4NetworkConfigurator to handle this though.

Update: I finally fix my local master. No more garbage commits.

### DIFF
--- a/src/inet/networklayer/configurator/ipv4/Ipv4NetworkConfigurator.cc
+++ b/src/inet/networklayer/configurator/ipv4/Ipv4NetworkConfigurator.cc
@@ -80,6 +80,32 @@ void Ipv4NetworkConfigurator::initialize(int stage)
         dumpConfiguration();
 }
 
+void Ipv4NetworkConfigurator::handleMessage(cMessage *msg)
+{
+    if(msg->isSelfMessage()) {
+        if(msg->getKind() == MCAST_JOIN_TIMER) {
+            InterfaceInfo::contextPtr_t *joinContextPtr = (InterfaceInfo::contextPtr_t *)msg->getContextPointer();
+            ASSERT(joinContextPtr);
+            if(joinContextPtr->interfaceData)
+                joinContextPtr->interfaceData->joinMulticastGroup(joinContextPtr->multicastGroup);
+            free(joinContextPtr);
+            cancelEvent(msg);
+        }
+        else if(msg->getKind() == MCAST_LEAVE_TIMER) {
+            InterfaceInfo::contextPtr_t *leaveContextPtr = (InterfaceInfo::contextPtr_t *)msg->getContextPointer();
+            ASSERT(leaveContextPtr);
+            if(leaveContextPtr->interfaceData)
+                leaveContextPtr->interfaceData->leaveMulticastGroup(leaveContextPtr->multicastGroup);
+            free(leaveContextPtr);
+            cancelEvent(msg);
+        }
+        else
+            throw cRuntimeError("unsupported self message in Ipv4NetworkConfigurator.");
+    }
+    else
+        throw cRuntimeError("Ipv4NetworkConfigurator does not receive any external msg.");
+}
+
 void Ipv4NetworkConfigurator::computeConfiguration()
 {
     EV_INFO << "Computing static network configuration (addresses and routes).\n";
@@ -187,6 +213,8 @@ void Ipv4NetworkConfigurator::configureRoutingTable(IIpv4RoutingTable *routingTa
 
 void Ipv4NetworkConfigurator::configureInterface(InterfaceInfo *interfaceInfo)
 {
+    Enter_Method("Ipv4NetworkConfigurator::configureInterface");
+
     EV_DETAIL << "Configuring network interface " << interfaceInfo->getFullPath() << ".\n";
     InterfaceEntry *interfaceEntry = interfaceInfo->interfaceEntry;
     Ipv4InterfaceData *interfaceData = interfaceEntry->getProtocolData<Ipv4InterfaceData>();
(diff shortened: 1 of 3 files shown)


levy:
First, let me start with a techincal note. The way you are organizing your pull requests makes my life way more difficult than what is absolutely necessary. The problem is that you are merging master and often other branches into your pull requests, instead of simply rebasing them separately on the updated INET master branch. This method makes pull request difficult to understand, because they often contain several patches from several other branches. Merging such a pull request is not really an option, because now I would have to understand all the seemingly unrelated patches from the side branches. It would be much easier to add a pull request to INET if it were rebased on the INET master branch. This method would result in a much smaller list of patches strictly belonging to a feature.

As for the Ipv4NetworkConfigurator, it is not meant to handle dynamic configurations such as changing addresses, adding/removing routes, joining/leaving multicast groups, etc. There's a separate component for handling dynamic scenarios, it is called ScenarioManager. I understand that the network configuration is related to this module, but I'm not sure that we want to complicate the already quite complex static Ipv4NetworkConfigurator module with dynamic behavior. I'm open to discussion.

levy:
I admit that force pushing the master branch wasn't very nice of me, but as a consequence at least I'll be having fun for eternity in hell instead of being bored to my second death in the heavens. :-) Nevertheless, the master branch is pushed every now and then with new features. So, if you need anything from master in your PL, it's best to _rebase_ your PL instead of merging master into it.

As for the dynamic behavior of simulations (e.g. network configuration), what about having a much more flexible module which would provide Python scripting similarly to what ScenarioManager provides using XML? For example, you could call the joinMulticastGroup/leaveMulticastGroup functions at certain simulation times, use for loops, if statements, etc.

levy:
There's a new INET branch (topic/python-experiment) which adds a new experimental PythonScripting module. This module takes a script parameter, which allows the following configuration, for example:

> **.script = "\
> host1 = simulation.getNetworkNode('host1');\
> eth0 = host1.getNetworkInterface('eth0');\
> at(1, lambda: eth0.joinMulticastGroup('224.0.0.7'));\
> "

This configuration adds the specified multicast group to eth0 of host1 at simulation time 1. Of course, there are other possiblities, we plan to generate the C++ binding automatically for all INET modules. This would allow using any public API from Python scripts. The end result is that users could easily implement complicated dynamic scenarios (including creating modules, initiating lifecycle operations, changing configurations, etc.) using the full power of Python (methods, control structures, etc.) What do you think about this approach?

ManiAm:
>> ... of being bored to my second death in the heavens. :-)

I suggest you to buy a play station :)

>> What do you think about this approach?

It is a powerful approach indeed and gives much more flexibility to scenarios specially in vehicular scenarios that I was working in VENTOS framework. However, don't you think the following syntax is much more user-friendly, cleaner and easier to understand for multicast address assignment?

`<multicast-group hosts='host0' address='224.0.1.1' joinTime='5' leaveTime='20' />`

IPv4 assignment is the first task a user performs for nodes in a network, and I personally prefer one place to assign all my IPv4 addresses (unicast and multicast). I do not want to write a python script for multicast addresses. The python script is very powerful for more complex scenarios however I prefer sticking to the old Ipv4NetworkConfigurator module.

Moreover, IP address assignment (unicast and multicast) should be coherent. It is not desirable to move from the following configuration for IPv4 assignment, for example,

```
*.configurator.config = xml("<config> \
			<interface hosts='**' address='x.x.x.x' netmask='255.x.x.x'/> \
  			<multicast-group hosts='host1' address='224.0.1.1' joinTime='5' leaveTime='20' /> \
       </config>")
```

To these :)

```
*.configurator.config = xml("<config> \
			<interface hosts='**' address='x.x.x.x' netmask='255.x.x.x'/> \
       </config>")
```

```
**.script = "
host1 = simulation.getNetworkNode('host1');
eth0 = host1.getNetworkInterface('eth0');
at(1, lambda: eth0.joinMulticastGroup('224.0.1.1'));
// python code for leaving
"
```

I agree that a user might need a more complex behavior in some scenarios. For example, host1 joins multicast group X1 at time T1, then joins to X2 at time T2, then leave X1 at time T3, etc. Such complex scenarios can be performed using the python script. This is very similar to the `find` dialog window in the MS Word. It is simple and has a minimal user interface and 90% of people are happy with it. If you want more advanced features, you click advanced to get access to more advanced options.

So in summary, I would add very basic multicast join/leave feature in the Ipv4NetworkConfigurator module that is most probably enough for many users. If a user needs more advanced control, then they can use the Python code.

levy:
>     <multicast-group hosts='host0' address='224.0.1.1' joinTime='5' leaveTime='20' />

If we are talking about the simplicity of syntax (XML vs Python), then we should compare apples to apples. For me, this is equally simple:

     configureMulticastGroup(hosts='host0', address='224.0.1.1', joinTime=5, leaveTime=20)
The above is valid Python code.

Anyway, I understand your reasoning, namely that you want to be able to configure simple things simply, and also related things together, but for me it's not strong enough. What I strongly do not agree with is to add dynamic configuration capabilities to Ipv4NetworkConfigurator, which is a static configurator by design. My reasoning is that there are zillions of different dynamic configuration possibilities, and they won't combine well in a simple XML file. Do you also plan to add time parameters to the manual router configuration entries of IPv4NetworkConfigurator? What about configuring the IP addresses over time? What about changing static autorouting at a specific time? And so on... We will end up with reinventing the wheel, a programming language in XML, and/or covering a very narrow set of example configurations.

If dynamic configuration is to be implemented in XML, which I think is a terrible idea, then I would still put it into ScenarioManager, possibly with the help of other modules (e.g. InterfaceEntry). For example:

    <join-multicast-group t='5' hosts='host0' address='224.0.1.1'/>
    <leave-multicast-group t='20' hosts='host0' address='224.0.1.1'/>

My reasoning is that what if you want to do this in a loop, or do it conditionally, etc.? Dynamic behavior is pretty much like using a programming language.

levy:
I forgot to mention another design consideration regarding Ipv4NetworkConfigurator. It doesn't actually configure IP addresses and routes, it might do other things directly but that's just laziness. Ipv4NetworkConfigurator just simply computes the configuration and stores it for future use. It's up to the network nodes to actually ask for specific configuration and use it. The reason is that network nodes may shut down and forget their configuration, but when they startup again, they are supposed to reconfigure based on the previously computed configuration. This way Ipv4NetworkConfigurator acts pretty much like a book keeper of network configuration.

levy:
I close this PR because according to the above discussion it's not a change in the right direction. The feature would be nice but if we are going to implement it then we have to come up with a different solution.

PR #400, opened 2018-11-30
Broadcast

ManiAm:
We can now selectively enable/disable direct broadcast on each interface. By default, direct broadcast is off on all interfaces due to security reasons (similar to almost all implementations). This PR does not affect the existing fingerprints except `examples/inet/broadcast` that is also updated in this PL.

For a full example, you can refer to the following:

https://github.com/ManiAm/inet/blob/INETex3/examples_INETex/multicasting/omnetpp.ini


levy:
Thanks!

PR #403, opened 2018-12-13
Add signal arrival time to the incoming packets

tim-ist:
This patch adds a tag carrying the signal arrival time to packets received by the radio.
It is useful when implementing TDMA protocols as it allows the MAC layer get the time of the packet arrival and synchronise precisely with the time slot structure.

This is related to my [question](https://stackoverflow.com/questions/53394411/get-the-time-of-arrival-for-a-wireless-frame) on StackOverflow.


levy:
Thanks! I'll update this change slightly to include the end time too.

PR #405, opened 2019-01-06
Proxy ARP

ManiAm:
Proxy ARP is a feature of ARP and needs to be defined in the ARP module and not IPv4. By default, proxy ARP is `true` in routers and `false` in hosts.

[commit 1498038]

In most implementations, proxy ARP can be enabled/disabled per interface. This commit adds this feature to the ARP operation.

Examples:

https://github.com/ManiAm/inet/blob/INETex3/examples_INETex/address_resolution/omnetpp.ini


ZoltanBojthe:
thank you

PR #430, opened 2019-04-13
fix addUdest method

Milad321:


rhornig:
This PR does not merge correctly, fails the integration test, unclear what it does and brings a lot of unrelated changes. Please send only concise PRs with reasoning, why it is needed and what it solves.

PR #452, opened 2019-07-22
BGPv4 additional implementation

kjerabek:
This commit contains BGPv4 implementation. It extends current functionality
of BGP in INET. It covers implementation of the following problems:

- BGP Update can carry more than one prefix.
- ConnectRetryTimer fix.
- Not consistent FSM states fix.
- IPv6 protocol support.
- Connection failure simulation support.
- Connection failure recovery simulation support.
- Multipoint connection prefixes and AS in update fixed.
- Internal and External neighborship behavior fixed.

New BGP examples were added to examples/bgpv4, the old ones were moved to
examples/bgpv4/oldExamples.


rhornig:
Hi Kamil,
thanks a lot for your contribution to the INET BGP protocol. We quickly went through the code changes and ran into some questions that should be clarified to get the PR reviewed and merged into the main INET repo. We are trying to keep the INET codebase stability as much as possible so we always try to keep the code changes in a single commit at manageable chunks. Right now the PR contains a lot of (somewhat) unrelated changes and it's really hard to see at first which changes belong to which functionality. To get it merged this PR should be broken down into smaller commits that  implement the various features you mentioned in the PRs description and we should work on / review these individually. 

Right now I'm seeing that the BGP examples were moved into an other directory and I'm not sure why is that justified/needed. I saw also several times that an existing API method was removed/commented out without indication exactly why? As we want to keep compatibility as much as possible any change to old APIs should be justified. Also the PR needs to be re-based on top of the current master branch so we can run the automated tests to see whether there are no regressions where current tests broke down.
We usually keep documentation changes/refactorings (that does not change the code behavior in any way) in separate commits.

There is a bit more info on how to prepare a PR that would immensely speed up the review process:
https://inet.omnetpp.org/SubmittingCode.html

I'm not sure how much time/resources you have for contributing back this code, but we would prefer to receive several smaller PRs so we can review/merge them faster.

kvetak:
Hello, Rudolf and Levente!

We were probably too hasty with the pull request. But please be patient with us, since it is PR done by us (from ANSA to INET) and we are walking this path for the first time. 

The original source code comes from my master student Adrian Novak (@xnovak1j). Kamil (or Marcel @screw) is going to do dirty work on this PR in order to prepare it for smooth acceptance. Rudolf, we will try to address most of your comments mentioned above in upcoming weeks and probe you again soon.

This PR follows the OMNeT++ Summit paper about BGP [OMNeT_Summit_2019__Vykosteny_.pdf](https://github.com/inet-framework/inet/files/3440940/OMNeT_Summit_2019__Vykosteny_.pdf). We are also cooking another PR targeting OSPFv3 (by Lukas Galbicka @Galino). We would like to discuss both of PR in the frame of the hackathon and coordinate with you more any future contributions to INET.

Best regards to everyone involved,

Vladimir

avarga:
Hi Vladimir,
Thank you for the background info and for the reference to the paper! If you have an update to the pull request the paper will help us better understand the changes, and indeed the Hackathon will be an excellent opportunity to discuss the remaining issues. Looking forward to meeting you there!

ZoltanBojthe:
> The BGP pull request should be withdrawn by Kamil because it needs a serious rewrite to be compatible with INET 4.1. _(Vladimír Veselý)_

PR #457, opened 2019-09-04
OSPF version 3

screw:
Here is our contribution of protocol OSPF version 3.
It was developed by Lukas Galbicka @Galino and adapted by me.
It should adhere to most of the INET contribution guidelines, but we'll await further instructions.
Hopefully, it will be able to merge it without any major changes.

We are aware of the following limitations:

P0 - Shortest Path First is missing
	- P0.2 - routing in P2P type of network
	- P0.3 - routing in NBMA type of network
	- P0.4 - routing in P2MP type of nwtwork
	- P0.6 - inter-AS routing
	- P0.7 - routing through virtual links

P1 - TODO:	Osfpv3Area.cc:755 - Link State ID  depends on the number of originated Router-LSAs by this process. For now, there is always only one Router-LSA from one process.

P6 - Aging of Link-State Database is not implemented. (solved)
	- TODO:	Ospfv3Area.cc:206, Ospfv3Interface:149 - unreachability is not managed, Should be on places where it is as a comment
			Ospfv3Adrea.cc:730 - Protocol creates only some LSA. After new LSA will be added, ageing of those needs to be implemented.

P8 - Options is not set in any LSA message
	- TODO:	Ospfv3Area.cc:761 - LSA Options for RouterLSA is not set
			Ospfv3Area.cc:1081 - LSA Options for NetworkLSA is not set
			Ospfv3Interface.cc:269 - Options for Hello Packet is not set.

P12 - LSR miss the calcualtion of:
	P12.1 - TTL
	P12.2 - Checksum

P13 - Protocol can not resend lost LSA messages of type:
	P13.1 - 3
	P13.2 - 4
	P13.3 - 5
	P13.4 - 9

P24 - In case of P2P type of connection routers get stuck in EXSTART state.

P27 - Input config file checking is not complete. In case of a wrong config file, an error can occur.

P28 - Metric calculation for inter-area routing is not always correct.

P30 - OSPFv3 Load Balancing is missing.


ZoltanBojthe:
Merged to INET master

PR #459, opened 2019-09-15
Improvements to STP/RSTP

ManiAm:
Examples are here:

https://github.com/ManiAm/inet/tree/INETex/examples_INETex/STP

### DIFF
--- a/src/inet/linklayer/configurator/Ieee8021dInterfaceData.cc
+++ b/src/inet/linklayer/configurator/Ieee8021dInterfaceData.cc
@@ -56,6 +56,7 @@ std::string Ieee8021dInterfaceData::detailedInfo() const
     out << " state:" << getStateName();
     out << " priority:" << getPortPriority();
     out << " linkCost:" << getLinkCost();
+    out << " isEdge:" << isEdge();
 
     return out.str();
 }
--- a/src/inet/linklayer/configurator/Ieee8021dInterfaceData.cc
+++ b/src/inet/linklayer/configurator/Ieee8021dInterfaceData.cc
@@ -64,43 +64,46 @@ std::string Ieee8021dInterfaceData::detailedInfo() const
 const char *Ieee8021dInterfaceData::getRoleName(PortRole role)
 {
     switch (role) {
-        case ALTERNATE:
-            return "ALTERNATE";
+    case NOTASSIGNED:
+        return "NOTASSIGNED";
 
-        case NOTASSIGNED:
-            return "NOTASSIGNED";
+    case ROOT:
+        return "ROOT";
 
-        case DISABLED:
-            return "DISABLED";
+    case DESIGNATED:
+        return "DESIGNATED";
 
-        case DESIGNATED:
-            return "DESIGNATED";
+    case ALTERNATE:
+        return "ALTERNATE";
 
-        case BACKUP:
-            return "BACKUP";
+    case DISABLED:
+        return "DISABLED";
 
-        case ROOT:
-            return "ROOT";
+    case BACKUP:
+        return "BACKUP";
 
-        default:
-            throw cRuntimeError("Unknown port role %d", role);
+    default:
+        throw cRuntimeError("Unknown port role %d", role);
     }
 }
 
 const char *Ieee8021dInterfaceData::getStateName(PortState state)
 {
     switch (state) {
-        case DISCARDING:
-            return "DISCARDING";
+    case BLOCKING:
+        return "BLOCKING";
 
-        case LEARNING:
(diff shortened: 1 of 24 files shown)


ZoltanBojthe:
I try to run fingerprint tests for STP/RSTP simulations: (on topic/ManiAm-stp branch in inet repo)

These simulations generates errors:
- `/examples/ieee8021d/ -f omnetpp.ini -c STP-LargeNet -r 0  (runtime error with exitcode=1: <!> Error: Logic error: frame BPDU from higher layer has local MAC address as dest (AA-AA-AA-00-00-02) -- in module (inet::EtherMac) LargeNet.switch2.eth[3].mac (id=139), at t=79.000001348s, event #57894)`
- `/examples/ieee8021d/ -f omnetpp.ini -c STP-LinkReconnect -r 0  (runtime error with exitcode=1: <!> Error: Logic error: frame BPDU from higher layer has local MAC address as dest (AA-AA-AA-00-00-03) -- in module (inet::EtherMac) LargeNet.switch3.eth[2].mac (id=178), at t=79.000001348s, event #59736)`

All other STP/RSTP simulation reports FAILED fingerprint values

ManiAm:
Ok, I am going to take a look at it.

ManiAm:
Still haven't found any free time to look into this.

I think the failed fingerprint is expected. In the original INET code, each switch is using the MAC address of its first Ethernet interface as its Switch MAC. I do not know what is the logic/rationale behind this, but in almost all implementations, each switch has a base MAC port that is unique. When you power up a switch, you will see its base MAC address getting printed on the console. In my PL, I have added the base MAC port concept and each switch is using it for communication. Hence the change in fingerprint.

ZoltanBojthe:
fingerprints calculated from these values of events:
  t - time of event
  p - full path of event handler module
  l - length of handled cPacket
  x - extra values added with FINGERPRINT_ADD_EXTRA_DATA() or FINGERPRINT_ADD_EXTRA_DATA2() in c++ code

avarga:
Just talked to Zoltan, what he meant was that actual frame content doesn't affect the fingerprint, so the fact that bridgeAddress is now computed differently shouldn't make a fingerprint test fail (unless the values actually affect the communication or the outcome of the algorithm somehow). So maybe there is some other issue. We'll continue next week by testing which commit changes the fingerprints.

ManiAm:
Thanks for the clarification Andras. Then I need to sit and check where the problem is!
After yesterday's changes to STP/RSTP, probably I need to invest even more time :)

avarga:
Thanks for this, @ManiAm — and apologies it sat open so long. I finally went through it carefully against current master.

The honest summary is that STP/RSTP has been reworked fairly extensively in the years since you opened this, so most of the PR has been overtaken by independent work — usually arriving at the same destination you were aiming for. Rather than leave it lingering, I'd like to close it as superseded, with credit where it's due. Here's the mapping, for the record.

Already in master (done independently, often along the lines you proposed)

- BPDU serializer — there's now an Ieee8021dBpduSerializer for byte-level (de)serialization, same idea as yours.
- BPDU message cleanup + named constants — the message was restructured into a BpduBase / BpduTcn / BpduCfg class hierarchy with BpduType / BpduProtocolVersionIdentifier enums, replacing the old magic numbers - the same intent as your struct grouping and enums.
- STP vs RSTP port states — the BLOCKING-vs-DISCARDING distinction you added is in, and extended further (BLOCKING + LISTENING for STP).
- WATCHes / debug visibility — added across the board.
- Plus a lot that goes beyond the PR: a real RSTP Proposal/Agreement handshake, TC-storm fixes, an 802.1D hold timer, and related cleanups.

Not adopted (deliberately)

- Base-MAC-address bridge identifier (getBaseMacAddress() and switching the bridge ID off it) — this is the change behind the runtime error and fingerprint failures Zoltán reported in this very thread back in 2019. Deriving the bridge address from the first non-loopback interface is retained on purpose, so this part is best left out.
- Renaming the XML attributes cost→portCost / priority→portPriority - backward-incompatible for existing configs, without a corresponding benefit.

Worth keeping — being extracted with attribution

- Your datarate-based default port cost (IEEE 802.1D-2004, Table 17-3) is a genuinely nice improvement that has no equivalent on master — today the configurator still uses a flat default. I'm lifting that idea into the current L2NetworkConfigurator as a separate, properly-tested change (re-implemented against the current code, since the file and surrounding APIs have moved), and you'll be credited for it.

So I'm closing this as superseded. Note that because none of the commits landed verbatim, GitHub won't mark it as merged. Thanks again for the contribution and for your patience — the port-cost idea in particular will make it in.

PR #475, opened 2019-10-28
TCP/IP: Added ECN for TCP/IP

m-habib:
This implementation supports only TCPReno and PPP. 
Implemented RFC-3168.

ECN is activated by setting the following in .ini file:
- Setting networkLayer.ip.ecnEnabled = true.
- Setting tcp.ecnWillingness = true for the TCP source and destination.
- Setting networkLayer.ip.ecnQLengthThrsh for the queue length threshold.

Added ECE and CWR flags to TCP header.

In TCP connection setup phase, the source and destination TCPs exchange information about their willingness to use ECN.

When routing a packet to the relevant output interface, if the interface's current queue length is above the specified threshold, IP layer sets the CE bit in the IP header.

For arriving packets with IP CE set, the receiver sends an ACK packet with the ECN-Echo (ECE) bit set. After sending the first packet with ECE set, the sender continues to set the ECE bit in all the ACK packets it sends until it receives a packet with CWR set.

For arriving packets with TCP ECE set, the sender reduces its cwnd and ssthresh and sends the next packet with CWR bit set (Congestion Window Reduced) in the TCP header. The sender will not react to congestion indications (ECE) more than once every rtt. The sender also will not send ECT packets (ECN capable) for re-transmitted packets.

Added an example of an ECN capable net.


src/inet/transportlayer/tcp/TCPConnectionUtil.cc
```diff
@@ -491,6 +531,30 @@ void TCPConnection::sendSynAck()
...
+    } else {
+        tcpseg->setEceBit(false);
+        tcpseg->setCwrBit(false);
+        if (state->endPointIsWillingECN)
+            EV << "non-ECN-setup SYN-ACK packet sent\n";
+    }
+    if (state->ecnWillingness && state->endPointIsWillingECN) {
+        state->ect = true;
+        EV << "both end-points are willing to use ECN... ECN is enabled\n";
+    } else { //TODO: not sure if we have to.
```
m-habib: I'm sure :-) 
Forgot to delete this comment

ZoltanBojthe:
TCP part merged to inet master.

IPv4 part rejected. Recommended to use the RedMarker module in the interface queue instead.

m-habib:
Hi Zoltan,

Thank you for the recommendation.

It was unclear in rfc-3168 whether packets CE marking should be made by RED or by IPv4. So I chose to let IPv4 access its own header and mark the packets.

Anyway, I've implemented CE marking in AlgorithmicDropperBase based on RED's decision to drop the packet, according to rfc-2884. And removed the packets marking from IPv4.

After the changes, ECN is activated simply by setting:
- tcp.ecnWillingness = true 

in the .ini file.

I've also added a new ECN Capable Net example.

Thanks!

ZoltanBojthe:
The ECN code in AlgorithmicDropperBase is not a good idea, it is a base class for all droppers.

The correct solution:
- adding ECN support for RedDropper,
- create new ethernet queue (e.g. EtherRedQueue) based on EtherQosQueue, add RedDropper with ECN support.
- using the new EtherRedQueue for ECN capable eth interfaces

ZoltanBojthe:
RedDropper with ECN support implemented, EtherQosRedQueue added in INET master.

m-habib:
Thanks, Zoltan,

Good luck!

levy:
We refactored and merged the ECN related code into the RedDropper module. The main reason for the change was to avoid code duplication between the old RedDropper and RedMarker modules and to keep the code as simple as possible. Since you were one of the people who brough the ECN support up originally, it would be good for us to hear what you think about the current solution. So do you have any comments on the current state of the ECN support in the INET master branch?

m-habib:
Hi Levente,

I will gladly take a look at the current solution in the INET master when I find free time (currently I'm in exam period). 
I also have some suggestions for ECN enhancements and extensions in both TCP and IP layers.

Will update as soon as possible :)

PR #488, opened 2019-11-30
Fix hard limit check on queue length on RED dropper

mgonzalezlopezudc:
Corrected bug in original code. The hard limit on the instantaneous queue length (queueLength) has to be checked independently of the conditions on the average queue length (avg). Otherwise, the hard limit is not respected when minth < avg < maxth and queueLength > maxth.


ZoltanBojthe:
fix added to master

PR #492, opened 2019-12-04
Pull ospfv2 nexthop

screw:
The next-hop being added has to have different Address (hopAddress)
or it has to be connected through different interface (ifIndex).
It allows to have two entries for the same next-hop if the next-hop(IP Address)
is reachable through multiple interfaces (further connected through
hubs/switches ...).


ZoltanBojthe:
Thanks, merged to inet master

Issue #497, opened 2019-12-18
"current connection stay listening.." TODO - already DONE?


screw:
It looks like the following TODO has been already DONE https://github.com/inet-framework/inet/blob/7a47b338fd92bfa520007086e4cca1067b2b6e60/src/inet/transportlayer/tcp/TcpConnectionRcvSegment.cc#L783-L788

which makes the comment on line 785 confusing
`TcpConnection *conn = cloneListeningConnection();    // "conn" is the clone which will stay LISTENing, while "this" gets updated with the remote address`
and contradicts this:
`EV_DETAIL << "Connection forked: new connection got new socketId=" << conn->socketId << ", "
                                                                                           "old connection keeps LISTENing with socketId=" << socketId << "\n";`

levy:
Pull request is merged.

PR #498, opened 2019-12-18
Tcp: Removed TODO; Updated comment

screw:
The TODO is already DONE;
Updated the comment to reflect that.
The clone handles the new connection and the (old) listening connection stays LISTENing.
#497


ZoltanBojthe:
Thank you, merged to master.

PR #503, opened 2020-01-03
Ieee802154Mac: remove unused radio signal supcription

ababila:
The signal subscription in the module seems to be not necessary


levy:
Thanks for the suggestion. The code has changed a bit since this PR was submitted, so I pushed an updated patched.

PR #504, opened 2020-01-03
Ieee802154Mac: statistic recording of the number of failed transmission attempts

ababila:
Frames are taken from the queue only if CCA successes, that is why the `currentTxFrame` has `nullptr` alway before the condition `if(currentTxFrame)` is verified for incrementing `nbDroppedFrame`. I think this is clearly an issue. 

Furthermore,  taking frames from the queue if and only if CCA successes imply never drop a frame in the head of the line of the queue until get transmitted, this simulates a queue of "infinite" length. I do not know if this was intended in the model, but I remarked that this affects positively packet loss rates on the link compared to this new approach (carried by this pull request) consisting of popping frames from the queue regardless if CCA will success or fail.


ababila:
This problem no longer appears in this commit 37e07d2b50f72cdada8d706f7addd69b5101b4da  on master.

PR #516, opened 2020-03-27
use deleteModule() to delete cSimpleModule

sommer:
Pull request for v3.6.7-bis.

This seems to be necessary at least since OMNeT++ 5.6 to keep deleteModule() of a parent module from failing.


riebl:
I believe no `Edcaf` submodule should be manually deleted at all because they will be destructed and deleted by the containing module (`Edca`).
In my opinion, `Edca`'s destructor should only `delete[] edcafs`, the manually allocated array of `Edca*` pointers.

sommer:
makes sense to me. I've amended the commit to simply remove deletion of the submodules

levy:
This problem is also present in the master branch.

PR #517, opened 2020-04-02
Enable configuration of recording transmission and/or reception in radio communication log files

ababila:
A feature that can optimize disk storage when transmission or reception traces are not needed in the log file.


levy:
Thanks for the PR, this is a good change, merged.

Issue #527, opened 2020-05-21
Several Modules incompatible with LifecycleOperation & patches to make them compatible


edlongman:
Several problems here but bear with me as I think some have simple solutions.

Setup 
----
`DymoRouter` with `Ieee802154NarrowbandMac` wireless interface. 
`SimpleEpEnergyStorage` and `SimpleEpEnergyManagement` operating a threshold on off system based on energy received from `AlternatingEpEnergyGenerator`

`wlan[0].radio` has `SensorStateBasedEpEnergyConsumer`

Network of 10 nodes with `UdpBasicApp` and `UdpSink` simulating "load" on network.

Problems and Patches
-------

* `Radio.cc` implements `handleStopOperation` and deletes any in progress transmissions on shutdown. `Ieee802154Mac` and presumably other MAC modules subsequently causes an unhandled null pointer Exception when accessing properties of `currentTxFrame`
https://github.com/inet-framework/inet/blob/76f495bca447551d7528c0dd5c153ecdebf94147/src/inet/linklayer/ieee802154/Ieee802154Mac.cc#L450-L455
`if (event == EV_FRAME_TRANSMITTED && currentTxFrame != nullptr)` 
Is enough to check that the currentTxFrame hasn't been deleted

* The `OperationalBase` picks up uncancelled selfMessages like `ccaTimer` and `rxAckTimer` after stopping. This throws a RuntimeError. 
How this should be fixed is up for discussion. Should the MacLayer implement `STOPPING_OPERATION` where it finishes the current transmission but then drops the rest of the queue? In a `CrashOperation` which is what the patch below is, the self messages should be cancelled by the MAC implementation where they are defined.
```
void Ieee802154Mac::handleStopOperation(LifecycleOperation *operation) {
    //TODO: More gracefully allow current Tx to end (delay operation)
    //Cancel all self message timers
    cancelEvent(backoffTimer);
    cancelEvent(ccaTimer);
    cancelEvent(sifsTimer);
    cancelEvent(rxAckTimer);
    MacProtocolBase::handleStopOperation(operation);
}

void Ieee802154Mac::handleCrashOperation(LifecycleOperation *operation) {
    cancelEvent(backoffTimer);
    cancelEvent(ccaTimer);
    cancelEvent(sifsTimer);
    cancelEvent(rxAckTimer);
    MacProtocolBase::handleCrashOperation(operation);
}
```

[...]

PR #531, opened 2020-05-30
Mark socket as CLOSED when closed locally and don't close unopened sockets #527

edlongman:
Alternatively, like TcpSocket. LOCALLY_CLOSED could be used.
UdpSockets can't be remotely closed anyway though.


ZoltanBojthe:
merged

PR #536, opened 2020-05-31
Topic/pull dctcp

screw:
This is a DCTCP implementation based on RFC 8257.

It based on the TCP Reno (slow start, fast retransmit, fast recovery, congestion avoidance).
In case of ECN (Explicit Congestion Notification) the DCTCP behaves differently than typical TCP.
Each ECN marked data packet is reflected in one Ack packet (normal TCP sends ECE marked Acks until it receives CWR). 

The sender then uses two equations to reduce cwnd
1. DCTCP.Alpha = DCTCP.Alpha * (1 - g) + g * M
where g is preconfigured variable and M is the ration of bytesMarked/bytesAcked.
and
2. cwnd = cwnd * (1 - DCTCP.Alpha / 2)

Also included is one scenario (Incast) with DcTcp config and TcpReno for easy comparison.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,3 +81,5 @@ backups
 *~
 xxx
 /tmp/
+/.computed/
+*.DS_Store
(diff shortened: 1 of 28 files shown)


ZoltanBojthe:
merged.

PR #537, opened 2020-06-05
Inetfix

aarizaq:
Several patchs for inet-framework

### DIFF
--- a/src/inet/common/Protocol.h
+++ b/src/inet/common/Protocol.h
@@ -133,6 +133,10 @@ class INET_API Protocol
     static const Protocol shortcutPhy;
     static const Protocol unitDisk;
     static const Protocol wiseRoute;
+
+    // Additional wireless protocols
+    static const Protocol lora;
+    static const Protocol lte;
 };
 
 inline std::ostream& operator << (std::ostream& o, const Protocol& t) { o << t.str(); return o; }
--- a/src/inet/common/misc/ThruputMeter.cc
+++ b/src/inet/common/misc/ThruputMeter.cc
@@ -46,7 +46,9 @@ void ThruputMeter::initialize()
 
 void ThruputMeter::handleMessage(cMessage *msg)
 {
-    updateStats(simTime(), PK(msg)->getBitLength());
+    auto pkt = dynamic_cast<cPacket *>(msg);
+    if (pkt)
+        updateStats(simTime(), pkt->getBitLength());
     send(msg, "out");
 }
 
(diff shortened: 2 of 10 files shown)


src/inet/common/Protocol.h
```diff
@@ -133,6 +133,10 @@ class INET_API Protocol
     static const Protocol shortcutPhy;
     static const Protocol unitDisk;
     static const Protocol wiseRoute;
+
```
levy, line 136: I guess these lines are added accidentally?

src/inet/common/misc/ThruputMeter.cc
```diff
@@ -46,7 +46,9 @@ void ThruputMeter::initialize()
 
 void ThruputMeter::handleMessage(cMessage *msg)
 {
-    updateStats(simTime(), PK(msg)->getBitLength());
+    auto pkt = dynamic_cast<cPacket *>(msg);
```
levy, line 49: Should use msg->isPacket() to check for being a packet, it's faster than dynamic_cast.

src/inet/physicallayer/ieee80211/packetlevel/Ieee80211ReceiverBase.cc
```diff
@@ -81,6 +81,7 @@ void Ieee80211ReceiverBase::setChannel(const Ieee80211Channel *channel)
         delete this->channel;
         this->channel = channel;
         setCenterFrequency(channel->getCenterFrequency());
+        setBandwidth(channel->getBand()->getSpacing());
```
levy, line 84: Band spacing is not the same as the listening bandwidth, so I think this is wrong.

aarizaq: It is true, but the channel structure only has information about the spacing and center, not the bandwidth, I have tried to include the minimum changes, without this modification it is impossible to use HT and VHT modes with bandwidths bigger than 20MHz. It should be necessary to include information about the bandwidth in the IIeee80211Band class. If you are busy, I can do it.

levy: Indeed, bandwidth should added to that interface and subclasses. You can give it a try, but please create a new pull request. This is unrelated to the others and it would be better to discuss and merge them separately.

aarizaq: Ok

src/inet/common/Protocol.h
aarizaq: Yes

src/inet/common/misc/ThruputMeter.cc
aarizaq: Ok, this simplify the if

aarizaq:
I have included a separate patch. My first approach was to extract the bandwidth directly from the modeset, but in this case the it is necessary to modify a lot of code, this is the simpler modification that in this moment I can imagine and that doesn’t introduce any conflict with the actual code and configuration parameters

levy:
I merged in the export fix part.

levy:
This PR has been obsoleted by a follow-up PR with different contents, so I'm closing it.

PR #539, opened 2020-06-06
Inetfix2

aarizaq:
Fix the bandwidth configuration problem in 5GH bands


levy:
I don't quite understand the reasoning behind this PR and the changes are also quite confusing. I'm closing this because in its current form this PR cannot be merged. The PR should clearly state a single problem and provide a solution, otherwise it's really hard to follow what's going on.

PR #544, opened 2020-06-26
queuing: Allow inverse priorities for PriorityScheduler

adriaan-niess:
For example in the IEEE 802.1Q standard's multi-queues, queue priorities are in descending order. So I think it makes sense to support inverse priorities in INET's PriorityScheduler module to be able to easily recreate such queuing networks while retaining the queue indizes given in these standards.


levy:
I agree, thanks!

PR #554, opened 2020-08-08
double-delete when both NetworkNodeOsgVisualizer and NetworkNodeCanvasVisualizer delete osg node?

sommer:
In INET 4.2.0, when deleting a module with a NetworkNodeOsgVisualizer at runtime, I am running into a double-delete that I don't quite understand - potentially because the `scene->removeChild` of https://github.com/inet-framework/inet/blob/v4.2.0/src/inet/visualizer/scene/NetworkNodeOsgVisualizer.cc#L85 also calls `delete`. Moving the `delete` from the base class to only the canvas visualizer works around this, but I'd love to get a second opinion on whether this is a bug to begin with.

### DIFF
--- a/src/inet/visualizer/base/NetworkNodeVisualizerBase.cc
+++ b/src/inet/visualizer/base/NetworkNodeVisualizerBase.cc
@@ -63,7 +63,6 @@ void NetworkNodeVisualizerBase::receiveSignal(cComponent *source, simsignal_t si
             if (isNetworkNode(module) && nodeFilter.matches(module)) {
                 auto visualization = getNetworkNodeVisualization(module);
                 removeNetworkNodeVisualization(visualization);
-                delete visualization;
             }
         }
     }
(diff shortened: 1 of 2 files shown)


levy:
Hmm, I'm not entirely sure but I think the scene->removeChild() call does not delete the child. I've checked the code and it seems to be the case, and it actually makes sense, because you could easily reuse the child somewhere else in the scene.

levy:
Also NetworkNodeCanvasVisualizer clearly doesn't delete the OSG node, because it has no access to it. So how can I reproduce this bug?

torokati44:
AFAIK OSG does automatic reference counting internally with `ref_ptr`, so a `removeChild()` may or may not delete the child Node, depending on its refcount. So, of there is a plain pointer to the Child somewhere (within or outside of the scene...), that might be a potential source of badness.

sommer:
It's been a while so I don't know if this triggers the described behavior, but following these steps seems to reproducibly crash the simulation

- use OMNeT++ 6.0pre9 
- use INET v4.2.1 
- Append to `examples/wireless/dynamic/omnetpp.ini` the following lines: 
``` 
*.visualizer.typename = IntegratedVisualizer 
*.visualizer.osgVisualizer.typename = "IntegratedOsgVisualizer" 
``` 
- execute this simulation in Qtenv (debug build in debug mode)
- switch to 3D 
- express run until the end 
- simulation crashes at https://github.com/inet-framework/inet/blob/v4.2.1/src/inet/visualizer/base/NetworkNodeVisualizerBase.cc#L66 with EXC_BAD_ACCESS

levy:
I was able to reproduce this bug and I also fixed it in master.

PR #556, opened 2020-08-18
Handle Ethernet frames in QoS classifiers

### LINKED ISSUE #548: QosClassifier does not work on wireless interface of access point
Hi,
we are currently looking into QoS in IEEE 802.11. We have a simple simulation model where UDP packets are exchanged between a wired and a wireless host via an access point.
Screenshot from 2020-07-13 14-48-05

We want to prioritize some of the traffic on the wireless network. For this, we enabled QoS and use the QosClassifier to classify the traffic. However, based on our simulation results, the classification and prioritization does only work for the wireless host and not for the access point. This can be seen in this screenshot:
Screenshot from 2020-07-13 15-20-00

Looking into details, we found out that packets on the access point have an Ethernet header when they hit the QosClassifier. However, the QosClassifier cannot handle the Ethernet header and thus classifies all traffic as AC_BE. For the wireless host, the packets do not have an Ethernet header and the classifier works like expected.

We modified the QosClassifier to handle the Ethernet header properly, which gives us the expected results, as shown in the screenshot below. Our modified version of the QosClassifier for INET 4.2.0 can be found in the attached archive together with our simulation model. The explained behavior has not changed with the current master version of INET. We are not sure if this is a bug or intended behavior, but are happy to provide a pull request with our fix for the current master version of INET.
Screenshot from 2020-07-13 15-24-29

Archive with simulation model and fixed QosClassifier:
[wiredandwirelesshostswithapqos.zip](https://github.com/inet-framework/inet/files/4912757/wiredandwirelesshostswithapqos.zip)

dstolpmann:
Hi,
I modified the QosClassifier and the ExampleQosClassifier so that they can also handle Ethernet frames. This fixes #548.

Best regards,
dstolpmann

### DIFF
--- a/src/inet/linklayer/common/ExampleQosClassifier.cc
+++ b/src/inet/linklayer/common/ExampleQosClassifier.cc
@@ -23,6 +23,9 @@
 #include "inet/linklayer/common/UserPriorityTag_m.h"
 #include "inet/networklayer/common/IpProtocolId_m.h"
 
+#ifdef WITH_ETHERNET
+#  include "inet/linklayer/ethernet/EtherFrame_m.h"
+#endif
 #ifdef WITH_IPv4
 #  include "inet/networklayer/ipv4/Ipv4Header_m.h"
 #  include "inet/networklayer/ipv4/IcmpHeader_m.h"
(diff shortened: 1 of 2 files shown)


ZoltanBojthe:
rebased, merged to inet master.

PR #557, opened 2020-08-19
fix: Prevent acos in antenna classes getting nan

TimonPetermann:
When simulating satellite networks, we observed the problem that the antenna gain is falling to 0 randomly due to small calculation errors in the dot product because of the large values in the position vectors. (getting absolute values slightly larger than 1)
This was only prevented in the AxiallySymmetricAntenna, thus for this pull request the implementation from the AxiallySymmetricAntenna is just copied to the other antenna classes where a acos is used.


levy:
Thanks!

PR #573, opened 2020-11-13
Bugfix: Segfault when accessing receptionTimer after endReception

edlongman:
This only occurs in very specific scenarios where:

* Reception has started
* The radio mode has begun switching, but not completed switching.
* The signal ends and `endReception` is called. But since the radio is now in the switching mode, the if statement on https://github.com/inet-framework/inet/blob/bbf2ef9cd8eb59271527cde549c5e9994f36df43/src/inet/physicallayer/wireless/common/radio/packetlevel/Radio.cc#L474 is false.
* The  timer object is deleted https://github.com/inet-framework/inet/blob/bbf2ef9cd8eb59271527cde549c5e9994f36df43/src/inet/physicallayer/wireless/common/radio/packetlevel/Radio.cc#L491 but `receptionTimer` variable still points to a deallocated section.
* The `completeRadioModeSwitch` calls `abortReception(receptionTimer)` which tries to access the deleted object.


levy:
Thanks for the PR!

PR #574, opened 2020-11-19
Report MTU when sending an ICMPv4 PTB

msvoelker:
* The ICMPv4 Packet Too Big message (type=3, code=4) includes the MTU as defined in RFC 1191.
* The new ICMP parameter quoteLength let the user specify the number of bytes to quote from the dropped packet.
* UDP forwards the ICMP packet inside the Control Info of the Error Indication to the upper layer protocol. This gives UDP based protocols access to the information in the ICMP packet.


ZoltanBojthe:
Merged to inet master, except the 'udp: let UDP forward the ICMP message to the upper layer' commit

PR #575, opened 2020-11-20
Bugfix to prevent wrong ...StateChangedSignal when Switching

edlongman:
Example that revealed problem:

1) Radio transmission from neighbour active on medium
2) Radio is commanded (e.g. from Mac layer) to receive mode, switching begins
3) Active transmission from neighbour ends and `endReception(timer)` is called
4) `updateTransceiverState()` determines `RECEPTION_STATE_BUSY` or `RECEPTION_STATE_IDLE`
5) Emits incorrect `receptionStateChangedSignal` and `transmissionStateChangedSignal`


levy:
Thank you for the PR and the good description!

PR #579, opened 2020-12-08
Clock: Add methods for rescheduling events

adriaan-niess:
In OMNeT++ 6.0 preview 9 the rescheduleAt and rescheduleAfter methods have been added to cSimpleModule. It might be a good idea to also add these methods to the ClockModuleMixin for a more consistent API.


levy:
Very nice, thanks!

PR #580, opened 2020-12-08
Fix InterpacketGapInserter

adriaan-niess:
Clock time is accessed before clock is initialized.


levy:
Hmm, testing with the IdealClock seems to be not so ideal after all. Thanks!

PR #630, opened 2020-12-14
RelayUnit: Consider user priority

adriaan-niess:
If we don't carry over user priorities (from IEEE802.1Q tag) in the relay unit, they'll get lost.


levy:
Makes perfect sense, thanks!

PR #650, opened 2021-02-09
Macos wifi fix

RedFu01:


levy:
Thanks for the PR. The related issue has been fixed in master with a different set of changes.

PR #672, opened 2021-04-08
VehicleMobility example and documentation

Lou8is:
Hi,

I wanted to try VehicleMobility and had trouble since it was lacking documentation and example, so here it is.


levy:
Thanks.

PR #692, opened 2021-05-06
Refactor out repeated code for matching Interface with Address

edlongman:
Nodes now require NextHopInterfaceData to be set, either with NextHopNetworkConfigurator or NextHopRoutingTable being used


ZoltanBojthe:
merged to master

PR #695, opened 2021-05-12
Work around omnetpp/omnetpp#867 by rounding down very small signal values

edlongman:
See  omnetpp/omnetpp#867 for long term fix to omnet but since the fix is not in omnet5 the work around is required


levy:
Isn't this obsolete now that the related bug has been fixed in omnet?

edlongman:
This will fix the inet version in omnet5.7 though, which will never receive the fix in omnet 6 I think.

levy:
I guess you meant *will not*. I think it's not worth the hassle to fix the older INET version because this issue must very rarely affect anyone. Don't you think?

edlongman:
I agree it's a rare issue.
Only really comes up in very large Sims and shouldn't be reported often
since omnet 6 will fix it

levy:
Then I close this.

PR #699, opened 2021-05-17
Explicitly inline some get module functions to prevent error when linking against INET api

edlongman:
Not sure exactly why this was a problem, and I think it is only a problem on Windows.

Caused by builiding another project that relies on INET and imports the dll. It will not link giving the following error: 
``` 
...
INET_API cGate *findConnectedGate(cGate *gate, int direction = 0)
                 ^~~~~~~~~~~~~~~~~
../../inet/src/inet/common/ModuleAccess.h:158:17: error: function 'omnetpp::cGate* inet::getConnectedGate(omnetpp::cGate*, int)' definition is marked dllimport
 INET_API cGate *getConnectedGate(cGate *gate, int direction = 0)
                 ^~~~~~~~~~~~~~~~
../../inet/src/inet/common/ModuleAccess.h:171:13: error: function 'T* inet::findConnectedModule(omnetpp::cGate*, int)' definition is marked dllimport
 INET_API T *findConnectedModule(cGate *gate, int direction = 0)
             ^~~~~~~~~~~~~~~~~~~
../../inet/src/inet/common/ModuleAccess.h:182:13: error: function 'T* inet::getConnectedModule(omnetpp::cGate*, int)' definition is marked dllimport
 INET_API T *getConnectedModule(cGate *gate, int direction = 0)```

Explicitly inlining these functions works though.

### DIFF
--- a/src/inet/common/ModuleAccess.h
+++ b/src/inet/common/ModuleAccess.h
@@ -128,7 +128,7 @@ INET_API InterfaceEntry *getContainingNicModule(const cModule *from);
  * Returns nullptr if no such module is found along the path.
  */
 template<typename T>
-INET_API cGate *findConnectedGate(cGate *gate, int direction = 0)
+INET_API inline cGate *findConnectedGate(cGate *gate, int direction = 0)
 {
     if (!gate->isConnectedOutside())
         return nullptr;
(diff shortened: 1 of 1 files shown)


edlongman:
Looks like this was already fixed actually in 5537bcd6773c6de1187e5e13c4771dc4b28c5507

Issue #700, opened 2021-05-17
`ScalarNoise::computeMaxPower` is incorrect when there is no background noise module


edlongman:
The way that compute max power and compute min power are calculated is not correct. 

For example: 
```
    simtime_t startTime = SimTime(10, SimTimeUnit::SIMTIME_MS);
    simtime_t endTime = SimTime(11, SimTimeUnit::SIMTIME_MS);
    inet::Hz centerFrequency = inet::MHz(1000);
    inet::Hz bandwidth = inet::kHz(1);
    std::map<simtime_t, inet::W> powerChangesBase;
    powerChangesBase.insert(std::pair<simtime_t, inet::W>(
            SimTime(5, SimTimeUnit::SIMTIME_MS), inet::W(5) ));
    powerChangesBase.insert(std::pair<simtime_t, inet::W>(
            SimTime(8, SimTimeUnit::SIMTIME_MS), inet::W(3) ));
    powerChangesBase.insert(std::pair<simtime_t, inet::W>(
            SimTime(18, SimTimeUnit::SIMTIME_MS), inet::W(-5) ));
    powerChangesBase.insert(std::pair<simtime_t, inet::W>(
            SimTime(20, SimTimeUnit::SIMTIME_MS), inet::W(-3) ));
    auto testPowerChanges = new std::map<simtime_t, inet::W>(powerChangesBase);
    inet::physicallayer::ScalarNoise test(startTime, endTime, centerFrequency, bandwidth, testPowerChanges);

    ASSERT(test.computeMaxPower(startTime, endTime) == inet::W(8) ); // Fails: result == W(0)
```

The result should clearly be `W(8)`. This happens whenever the "test interval" is shorter than any `powerChanges` gap, as well as some other boundary cases. If there is power change at a boundary change it works ok.

## Severity/Prevelance

* Is is common for such short "test intervals" because of `Radio::isListeningPossible()` : `const IListening *listening = receiver->createListening(this, now, now + 1E-12, position, position);`
* Most of the time `isListeningPossible` is called at the same time as receptionEnd or receptionStart so there is a powerChange `now`. So it still calculates this Ok
* If there is radio transition to `RECEIVER` or `TRANSCEIVER` during an ongoing radio signal, then there is no radio power change at `now` but there will be one from `...BackgroundNoise::computeNoise(...)`. Since the noise is modelled by a signal being added between listeningStart and listeningEnd. So it still calculates this Ok
[...]

PR #701, opened 2021-05-18
Improve ScalarNoise max and min computation

edlongman:
Fixes the behaviour for the computeMaxPower function in #700 

### Changes to out of range behaviour for `computeMinPower`
Previously a startTime and endTime out side of the range of power changes would return NaN. 

Now it returns 0 if out of range before and NaN if *both* are out of range after the power changes range

It is debatable what "correct" behaviour should be, this shouldn't ever actually be a problem though, looking at where the function is used. Since either 0 and NaN have the same effect to the subsequent computation *and/or* the powerChanges are exactly aligned with startTime or endTime


levy:
Thanks for reporting this issue and especially for also providing a test! I modified the test a bit to expect 0 when the range is after all the changes, and also committed a somewhat different fix.

PR #707, opened 2021-06-22
Remove redundant definition of IRadio and LayeredProtocolBase Signals

edlongman:


edlongman:
This doesn't do what I expected. It works in Release mode but doesn't in debug because the signal names are not explicitly redefined

PR #724, opened 2021-09-16
Add ModuleIdAddress and ModulePathAddress to get L3AddressResolver::findHostWithAddress

edlongman:
Created to replace #682 

Replaced the logic to check Ipv6, Ipv4 and Mac L3Address types with a call to the NetworkInterface::hasNetworkAddress(L3address ) which duplicated the logic.

Possibly the ModulePath and ModuleId addresses should also use the function too but it's unclear in the function description for findHostWithAddress whether it should exclusively look for Hosts with NetworkInterfaces with NextHopInterfaceData or whether it is sufficient for it to just be the node with the right module id or module path.

Currently in InterfaceTable::findInterfaceByAddress(L3Address ) it matches interfaces without NextHopInterfaceData


ZoltanBojthe:
Thanks for your fix!
Merged to inet master.

PR #726, opened 2021-09-27
Added current version of the HtbQueue

m-bosk:
Licensing potentially not final
HtbClassifier unchanged -> pending removal


levy:
I just merged this PR in as it is now. Additional changes can go into the topic/htb branch on top of this. Later we can squash the patches if needed.

PR #727, opened 2021-09-29
Fixed compatibility issues with OMNeT++ 6.0Pre12 and the current INET Version

m-bosk:


levy:
May I ask that you resubmit the pull request against the current topic/htb branch without the unnecessary master merge patches. Ideally we should have a linear branch that neither merges master nor merged into master. If you need any patch from master just cherry pick the patch into your branch and resubmit the pull request against the latest version available in the upstream repository. This approach makes it easier to merge your changes and also to understand what is going on.

m-bosk:
Will fix asap

PR #728, opened 2021-10-01
HTBQueue: Omnet6pre12 compatibility fix and remove the need for HtbClassifier

m-bosk:
Redoing the pull request after removing the accidental merge

Changes:
- Fixed HTB compatibility issues with OMNeT6.0pre12 and current INET version
- Removed the need for HtbClassifier. Any classifier can be used now instead. The default is ContentBasedClassifier
- Adjusted the HTB examples with the appropriate pattern matching syntax


levy:
Thanks for the changes. The change in the scheduler is not necessarily the way I would do it but it certainly a step in the right direction because it removed the specialized classifier. I will add the HTB queue specific fingerprint tests.

You will be able to run the tests as follows:

```
levy@valardohaeris:~/workspace/inet$ . setenv 
/home/levy/workspace/inet
levy@valardohaeris:~/workspace/inet$ cd tests/fingerprint/
levy@valardohaeris:~/workspace/inet/tests/fingerprint$ ./fingerprinttest --debug -m htb
```

levy:
So I added the HTB examples to the fingerpint tests in examples.csv. You may want to change the time limits though and update the fingerprints by running the above command and replacing the file with examples.csv.UPDATED. 

To get a good baseline for regression testing it would be nice to carefully check if the time limit is right and everything interesting happens within that limit and also that simulations run as expected.

PR #732, opened 2021-11-04
Fine-tuned HTB fingerprint tests and example simulations

m-bosk:
Sorry for no updates recently. We finally found some time to update the fingerprint tests with proper simulation times. Of course, we verified that all interesting things happen within the new, shorter, simulation time for the examples.

Note that we used the "~tND" argument instead of the "tyf" argument. The "Fake GUI" which got used originally with the "tyf" argument made the whole simulation fail on our machines. We were using Omnetpp-6.0pre12. That said, it works with the "~tND" argument.

Any feedback is welcome!

### DIFF
--- a/examples/htb/omnetpp.ini
+++ b/examples/htb/omnetpp.ini
@@ -48,7 +48,7 @@ record-eventlog = false
 [Config scenarioUDP1]
 description = "HTB Evaluation scenario I"
 network = htbEvaluation
-sim-time-limit=400s
+sim-time-limit=100s
 
 output-vector-file = "${resultdir}/${configname}/${configname}-${runnumber}.vec"
 output-scalar-file = "${resultdir}/${configname}/${configname}-${runnumber}.sca"
@@ -84,7 +84,7 @@ output-scalar-file = "${resultdir}/${configname}/${configname}-${runnumber}.sca"
 *.serverFDO.app[*].stopOperationTimeout  = 2s # timeout value for lifecycle stop operation
 
 *.serverFDO.app[*].startTime = replaceUnit(10*index+uniform(1,1.1), "s") # time first session begins
-*.serverFDO.app[*].stopTime = replaceUnit(10*index+uniform(100,100.1), "s") # time of finishing sending, negative values mean forever
+*.serverFDO.app[*].stopTime = replaceUnit(10*index+uniform(50,50.1), "s") # time of finishing sending, negative values mean forever
 
 *.nFDO =  1
 
@@ -116,7 +116,7 @@ output-scalar-file = "${resultdir}/${configname}/${configname}-${runnumber}.sca"
 [Config scenarioUDP2]
 description = "HTB Evaluation scenario II"
 network = htbEvaluation
-sim-time-limit=400s
+sim-time-limit=100s
 
 output-vector-file = "${resultdir}/${configname}/${configname}-${runnumber}.vec"
 output-scalar-file = "${resultdir}/${configname}/${configname}-${runnumber}.sca"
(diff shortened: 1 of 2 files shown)


levy:
The tyf fingerprint isn't that important, it's purpose is to test the appearance on the GUI, it can be re-added later.

levy:
I don't see any HTB specific classifier in the topic/htb branch anymore. I
guess it has already been removed, no?

There are still changes in the Ppp module though, which I will take care
of, so you can just ignore this.

Do you really need the fingerprint simulation time limits to be this large?
It takes a lot of time to complete the tests this way. We usually set the
time limits to the lowest possible value where all interesting (to be
tested) things already happen. The fingerprint tests run quite often so
it's desirable to make them fast.

Also, some of the fingerprint tests FAIL on my machine. I used
omnetpp-6.0pre12 to run the tests (and in DEBUG mode).

Failures:
  /examples/htb/ -f omnetpp.ini -c scenarioTCP1 -r 0  (some fingerprint
mismatch: actual '1a2c-182a/tplx;a987-b345/~tNl;6340-8ea3/~tND', expected:
'c489-5df7/tplx;1469-3bad/~tNl;e222-9b8f/~tND')
  /examples/htb/ -f omnetpp.ini -c scenarioTCP2 -r 0  (some fingerprint
mismatch: actual '4562-27ef/tplx;60da-62b3/~tNl;ec81-bec9/~tND', expected:
'd4cd-6af0/tplx;2feb-4705/~tNl;55f8-33e3/~tND')

Errors:
  /examples/htb/ -f omnetpp.ini -c scenarioUDP1 -r 0  (runtime error with
exitcode=1: <!> CPU time limit reached -- at t=46.072874985166s, event
#5640192)
  /examples/htb/ -f omnetpp.ini -c scenarioUDP2 -r 0  (runtime error with
exitcode=1: <!> CPU time limit reached -- at t=48.222682632497s, event
#6033408)
  /examples/htb/ -f omnetpp.ini -c scenarioUDP4 -r 0  (runtime error with
exitcode=1: <!> CPU time limit reached -- at t=45.703101090437s, event
#5500928)

The errors may go away if we would use a smaller value.

levy:
I tried in release mode and the errors turned into failures.

levy:
The std::set<htbClass*> fields may cause fingerprint instability because the key is a pointer value. The pointer value is quite random and depends on the memory allocation pattern and this may result in changing the order in the set. Iterating or even accessing the first entry of the set may become non-deterministic and cause fingerprint instability.

I quickly scanned through the code and found a few begin calls on those sets which are suspicious to me. You should either  change the key in the set to something else that is deterministic in the simulation or remove the order dependent access.

m-bosk:
Yes, we removed the classifier in one of the previous commits https://github.com/inet-framework/inet/pull/728#issue-1013193034. I believe you mentioned the last time that the removal we did was a step in the right direction, however it still might require some changes https://github.com/inet-framework/inet/pull/728#issuecomment-933245412.

We can shorten the fingerprint simulations even further. That shouldn't be an issue.

Regarding the randomness in the keys. Since we are using the class pointers themselves, it should be possible to change the keys to a deterministic value, i.e. the class name that already exists. We will look into that.

Other than that we are unsure why you got those errors. We'll do some more testing on our machines to get to the bottom of it.

levy:
You can ignore the errors, they are just timeouts on CPU time limit. The real issue now is non-determinism of the examples.

levy:
I'm sorry for the long delay.

The HTB queue/scheduler is merged into the INET master branch. We had to modify the module hierarchy with respect to where queues are placed inside the network interface. This change was not trivial and took quite a while but had to be done for other reasons too. The Ethernet HTB examples produce some errors that seems to be internal to the HTB scheduler modules. See #734 for more details.

PR #734, opened 2021-11-16
HTBQueue: Shorter fingerprint tests and fixed std::set determinism with pointers

m-bosk:
Shortened HTB fingerprint tests to at most 15 seconds each.

Also, added comparators to all uses of std::set and std::multiset to make them behave deterministically - i.e. each HTB class now has unique, deterministically assigned ID that is compared within the sets and multiset instead of the pointer itself.

A few words to the fingerprint tests. We have run the full suite on two separate machines multiple times and the tests were passing each time. So nothing changes between the runs. However, when the fingerprints are run on a new platform for the first time, some of them still fail. After an update to the fingerprints though, the tests pass every single time (at least for the multiple tries we did on a new machine). We are not 100% sure why that happens. We think that might have something to do with the platform dependencies...

### DIFF
--- a/examples/htb/omnetpp.ini
+++ b/examples/htb/omnetpp.ini
@@ -48,7 +48,7 @@ record-eventlog = false
 [Config scenarioUDP1]
 description = "HTB Evaluation scenario I"
 network = htbEvaluation
-sim-time-limit=100s
+sim-time-limit=11s
 
 output-vector-file = "${resultdir}/${configname}/${configname}-${runnumber}.vec"
 output-scalar-file = "${resultdir}/${configname}/${configname}-${runnumber}.sca"
@@ -75,7 +75,7 @@ output-scalar-file = "${resultdir}/${configname}/${configname}-${runnumber}.sca"
 *.serverFDO.app[*].destAddresses = "hostFDO[0]"
 *.serverFDO.app[*].destPort = 1042 + index
 *.serverFDO.app[*].messageLength = 1465B
-*.serverFDO.app[*].sendInterval = uniform(0.1ms, 0.11ms)
+*.serverFDO.app[*].sendInterval = 0.1ms
 *.serverFDO.app[*].packetName = "UDPData"
 *.serverFDO.app[*].typename = "UdpBasicApp"
 *.serverFDO.app[*].localAddress = "" # local address; may be left empty ("")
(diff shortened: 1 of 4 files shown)


levy:
My colleague @ZoltanBojthe took over the topc/htb branch and started working on the necessary changes in PPP in the topic/htb-bz branch. He will fix the platform dependency too and this branch will eventually be the topic/htb and also be squashed for easier review before merging it into master. The current version should run with omnetpp-pre14 (the latest). Please let us know if there's anything on your side. We will come back to you soon.

ZoltanBojthe:
On topic/htb-bz, I added the ScenarioUdp1Eth config for htb examples, for testing with ethernet interface.
Unfortunatelly, this example generates throw in HtbScheduler::htbDequeue() at second packet pulling. The canPullSomePacket() say true, because getNumPackets() > 0, but the CompoundPacketQueueBase::pullPacket() ---call---> HtbScheduler::schedulePacket() ---call---> HtbScheduler::htbDequeue() generates the throw: "The class leafhostFDO0 deficit on level 0 is negative!"
The original ScenarioUdp1 works without any error, but the first 3 packets sending without gaps, and from the 4th packet, the sending has big gaps between packets.

(for topic/htb-bz, currently required the  OMNeT++ 6.0 Preview 15)

ZoltanBojthe:
the inet/topic/htb-bz merged into inet master

m-bosk:
> On topic/htb-bz, I added the ScenarioUdp1Eth config for htb examples, for testing with ethernet interface. Unfortunatelly, this example generates throw in HtbScheduler::htbDequeue() at second packet pulling. The canPullSomePacket() say true, because getNumPackets() > 0, but the CompoundPacketQueueBase::pullPacket() ---call---> HtbScheduler::schedulePacket() ---call---> HtbScheduler::htbDequeue() generates the throw: "The class leafhostFDO0 deficit on level 0 is negative!" The original ScenarioUdp1 works without any error, but the first 3 packets sending without gaps, and from the 4th packet, the sending has big gaps between packets.
> 
> (for topic/htb-bz, currently required the OMNeT++ 6.0 Preview 15)

I just managed to have a very brief look at that and I believe that the error might actually be related to a misconfiguration of the HTB tree for Ethernet. I believe the issue is that the "quantum" parameter is set too low.

Reason: For PPP (as used in ScenarioUdp1) the maximum packet size is 1492 Byes. For Ethernet it is 1518 Bytes. The "quantum" parameter is closely related to the maximum packet size and `quantum >= maximum packet size` needs to hold. With quantum set to 1500 Bytes (as in the tree_scenario1.xml config), in Ethernet scenario, it could result in negative values for the deficit where the throw statement was placed. 

I will verify that and propose an appropriate change to the tree config beginning next week. 

Shortly regarding the first 3 packets: The HTB holds the set rates on average over time. Bursts are allowed and might happen.

levy:
Thanks for taking a look at it!

PR #739, opened 2022-01-25
HTBQueue: Fix & add examples working with Ethernet interface

m-bosk:
In HTBQueue Example with Ethernet interface, the `HtbScheduler::htbDequeue()` method was throwing the error: "The class leafhostFDO0 deficit on level 0 is negative!"

This error occurred due to the "quantum" value in the HTB tree being set too low. What happens when this value is too low, is that the "deficit" controlling the Deficit Round Robin part of the scheduling algorithm (DRR prevents flow starvation) could never recover above 0. "Quantum" essentially governs how the deficit is replenished after a packet transmission. It needs to be set to a value greater or equal the maximum packet size (including all headers, also PHY). The misconfiguration meant that the given class would not be able to be selected for dequeue. Hence the check for deficit < 0 that threw the error.

The quantum was set too low because the ethernet examples were based on PPP examples. Changing the interface type from PPP to Ethernet brings along a few necessary modifications. First, is a change in the quantum values ( in the xml tree configurations). Namely, the quantum value should be at least 1519 bytes, corresponding to the Ethernet MTU. In this fix we provide new configurations for the example with ethernet interfaces, including HTB trees with the quantum parameter being set slightly higher (to 1550 bytes) to allow for larger ethernet headers compared to ppp headers. Included are ethernet configurations for all previous examples. We did not touch the fingerprint tests.

Furthermore, in the current HTB implementation we only account for a fixed (hard-coded) physical layer header size. This value is needed to calculate the full packet size placed on the wire. We use the value for class deficit calculation as well as for token accounting in the nested token buckets. The physical layer header size is hard-coded to be 7 Bytes what corresponds to PPP PHY layer header size. As a different example, for ethernet this value should be 8 Bytes. A solution would be to implement a dynamic check that verifies what physical layer header size the used interface adds on top of the link-layer packet. We are currently unsure how to solve this. Any suggestions are welcome.

### DIFF
--- a/examples/htb/omnetpp.ini
+++ b/examples/htb/omnetpp.ini
@@ -158,7 +158,7 @@ output-scalar-file = "${resultdir}/${configname}/${configname}-${runnumber}.sca"
 *.hostFDO[0].eth[0].queue.numQueues = 5
 *.hostFDO[0].eth[0].queue.queue[*].typename = "DropTailQueue"
 *.hostFDO[0].eth[0].queue.htbHysterisis = false
-*.hostFDO[0].eth[0].queue.htbTreeConfig = xmldoc("tree_scenario1.xml")
+*.hostFDO[0].eth[0].queue.htbTreeConfig = xmldoc("tree_scenario1Eth.xml")
 *.hostFDO[0].eth[0].queue.queue[*].packetCapacity = 500
 *.hostFDO[0].eth[0].queue.classifier.defaultGateIndex = 1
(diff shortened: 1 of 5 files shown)


levy:
Thanks for the PR and the nice description. As for the PHY overhead, how about simply having a parameter in the scheduler? I guess the user can always set it properly, no? In general, it would almost be impossible to get this right automatically because "anything" can happen to the packet after the traffic shaper. So I guess what I'm saying is that there's no correct automatic solution for this kind of problem right now.

PR #740, opened 2022-02-04
HtbQueue: Dequeue DRR bugfix, corrected typos, and implemented PHY layer overhead parameter

m-bosk:
Fixed a bug where classes on higher levels would not be considered as candidates for dequeue in DRR:
- Classes of `level >= 1` while in inner feed of their parent would not get updated as candidates for next to dequeue.
- Caused classes that were not selected as initial next to dequeue on `level >= 1` to starve while in may_borrow mode.
- Bug caused by a typo in the comparison of classes. Instead of comparing the class on currently considered level in DRR, the corresponding leaf was always compared. That lead to problems if more than 2 levels were present and `ceil > rate` for inner nodes.

Implemented a parameter for physical layer overhead as discussed [here](https://github.com/inet-framework/inet/pull/739#issuecomment-1021235190).

Added fingerprint tests for configs with Ethernet.

Fixed a few other typos in the code.


levy:
Thanks for the PR, it looks ok to me. I merged this into master. I have a question though not related to this PR.

Due to complicated licensing issues, we decided that the next INET release will exclusively use the LGPL license, that is no GPL licensed component will be present in the main INET repository. Currently the HTB modules are licensed under GPL. Would it be possible to change the license to LGPL or is it not allowed because too much of the Linux kernel code is used in it?

If the license cannot be changed to LGPL, then the HTB modules will be moved to a separate INET-GPL repository before the next release. This other repository will contain all the other GPL licensed INET components and it cannot be distributed in binary form because it is linked to omnet libraries which are not GPL compatible.

PR #753, opened 2022-04-07
allow arbitrary wireless interface names (default: wlan cellular)

wischhof:
A minor extension of the `L3NetworkConfiguratorBase` class to allow other wireless interface names than "_wlan_".

Background: Currently, network layer configurators such as the `IPv4NetworkConfigurator` check if an interface is a wirless interface by comparing the interface name: if the interface name starts with "wlan", it is assumed to be a wireless interface. (see https://github.com/inet-framework/inet/blob/98e3ec7fc076fc4d974fa28a82092cc318a9966e/src/inet/networklayer/configurator/base/L3NetworkConfiguratorBase.cc#L269) 

As a result, a wireless interface name has to start with "wlan". For some of our simulations, where we combine wlan interfaces (from inet) with cellular network interfaces (from simulte/simu5g), this had the ugly consequence that wlan0 could be a wlan and wlan1 a LteNic. Since this is misleading, it would be great if we could get this small change (which should have no side-effects) upstream.

BTW: Of course, we can (and do) extend the class ourselves to get the desired behavior - however, from our point of view having this upstream would be a much cleaner implementation since it is then in the base class, where it belongs.

Any comments welcome ;-)

Lars

### DIFF
--- a/src/inet/networklayer/configurator/base/L3NetworkConfiguratorBase.cc
+++ b/src/inet/networklayer/configurator/base/L3NetworkConfiguratorBase.cc
@@ -57,6 +57,12 @@ void L3NetworkConfiguratorBase::initialize(int stage)
         minLinkWeight = par("minLinkWeight");
         configureIsolatedNetworksSeparatly = par("configureIsolatedNetworksSeparatly").boolValue();
         configuration = par("config");
+
+        std::string wirelessIfStr = par("wirelessInterfaces").stdstringValue();
+        cStringTokenizer tokenizer(wirelessIfStr.c_str(), " ,;\t\n\r\f");
+        for(const auto& s : tokenizer.asVector()){
+            wirelessInterfaceNames.push_back(std::make_pair(s, s.size()));
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


aarizaq:
you can use the gate radioIn also

bool L3NetworkConfiguratorBase::isWirelessInterface(NetworkInterface *networkInterface)
{
    return !strncmp(networkInterface->getInterfaceName(), "wlan", 4) || networkInterface->hasGate("radioIn");
}

wischhof:
> you can use the gate radioIn also
> 
> bool L3NetworkConfiguratorBase::isWirelessInterface(NetworkInterface *networkInterface) { return !strncmp(networkInterface->getInterfaceName(), "wlan", 4) || networkInterface->hasGate("radioIn"); }

Interesting suggestion - would also be a "minimal modification" that would work for us, you are right!

However, personally I think it is a cleaner approach to base the "detection" if an interface is wireless on a single criteria (->currently the name) instead of looking at the internal structure within the wireless interface.  But that's just my two cents ;-)

levy:
In the master branch there's already an isWired()/isWireless() method in the NetworkInterface class.

wischhof:
> In the master branch there's already an isWired()/isWireless() method in the NetworkInterface class.

That's great! - I did not know that. Then, I would suggest to replace all calls to isWirelessInterface(...) and remove the isWirelessInterface(..) method of L3NetworkConfiguratorBase. I will do that and open a separate, new pull request - so that the other commits are not included, ok?

wischhof:
Opened a new pull request #754 

Since this pull request is obsolete with that approach, I will close.

PR #770, opened 2022-06-15
Fix missing `INET_API` in nested `NetworkInterface::LocalGate` with vtable

edlongman:
Creating a custom network interface in an inet dependant project causes a linking error in `lld` on Windows.

This is because the vtable for the `NetworkInterface::LocalGate` is not exported into the DLL symbols list.


edlongman:
Thanks for the quick accept, 
I've done a quick search through the inet source with the following `\r\n[ \t]+class\s(?!INET)` and it's picked up a 144 other potential places where this might be a problem if INET base classes are used. 

Not a problem for me, but could be in the future for others trying to extend INET components, particularly: 
* `StatisticVisualizer`
* `...Antenna`
* `Arp` (`::ArpCacheEntry`)
* `L2NetworkConfigurator`

PR #772, opened 2022-06-21
Added  Dijkstra's algorithm to find shortest paths from a src to all other nodes

dragos-bth:
…other nodes

### DIFF
--- a/src/inet/common/Topology.cc
+++ b/src/inet/common/Topology.cc
@@ -466,6 +466,71 @@ void Topology::calculateWeightedSingleShortestPathsTo(Node *_target) const
     }
 }
 
+void Topology::calculateWeightedSingleShortestPathsFrom(Node *_target) const
+{
+    if (!_target)
+        throw cRuntimeError(this, "..ShortestPathTo(): target node is nullptr");
+    auto target = _target;
+
+    // clean path infos
+    for (auto& elem : nodes) {
+        elem->dist = INFINITY;
+        elem->outPaths.clear();
+    }
+
+    target->dist = 0;
+
+    std::list<Node *> q;
+
+    q.push_back(target);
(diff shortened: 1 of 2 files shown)


levy:
I have created a topic/topology branch with a couple of patches which contains your changes, but in this branch the two methods are merged into one. I tried to make the patches so that it is easy to follow the changes. I think it looks better this way because there's not much redundancy left. One thing I'm not sure is line 468 in Toplogy.cc:

```
            else if (!contains(remote->outPaths, to ? current->inLinks[i] : current->outLinks[i]))
                (to ? remote : current)->outPaths.push_back(to ? current->inLinks[i] : current->outLinks[i]);
```

Namely the expression (to ? remote : current) but I had to add this in order to match your version. I was doing this in a quite mechanical way. Is it required because we are storing the result in the opposite direction?

Anyway, could you please check if this solution satisfies your requirements?

As for your other note, I think we should address it in a separate issue.

levy:
Generally, we try to avoid having merge commits in the master branch to make following changes easier. So I just rebased the topic branch on top of the master branch and pushed upstream. Thanks for verifying!

dragos-bth:
Hi Levente,

I see that this change (issue #772) didn’t make it into the official v4.4.1 (mainline). Is there anything I can do to push it into the next release?

Best regards,
Dragos

rhornig:
Just to clarify: v4.4.1 is NOT the mainline (master) branch. That is a bugfix branch and we do not put new features (and especially API changes into patch releases). The PR changes was merged (manually) into the master branch. which will eventually be released as INET 4.5

dragos-bth:
Thank you for letting me know. I was not aware that 4.4.1 is only a bugfix branch.

Issue #771, opened 2022-06-20
[TSN feature] GCL capacity in GateScheduleConfiguratorBase or PeriodicGate


yue2388253:
Hi all,

The number of a gate control list is limited in a real TSN switch (e.g., 1024 gate control entries).
Current implementation does not take this limitation into consideration. 
Is it possible to implement this feature?

Best regards,

Lam

levy:
The gate schedule configurator modules configure all existing gates in the network. Obviously, there's no way to configure a gate that doesn't exist, so all it could do perhaps is not to configure some of the gates which are there. I don't know if that makes sense, because then how could the configurator make sure that the jitter and latency limits are met.

This feature doesn't make too much sense to me. Can you describe the use case you are trying to solve?

yue2388253:
The gate state is controlled by a gate control list (GCL) according to IEEE 802.1Qbv standard[1], as shown below.
![image](https://user-images.githubusercontent.com/33801130/174596646-09e7cb3f-f735-4f7e-b3b8-d847b51c03bd.png)

The length of a GCL is not inifinte, due to the limited hardware capacity. In fact, a typical TSN switch can only support 1024 gate control entries. Therefore, a GateConfigurator should take this GCL limitation into consideration, otherwise the calculated schedule would be impractical.

I think it would be great to set a upper bound for the GCL length and verify whether the output of a GateConfigurator is feasible.

Reference:
[1] "IEEE Standard for Local and metropolitan area networks -- Bridges and Bridged Networks - Amendment 25: Enhancements for Scheduled Traffic," in IEEE Std 802.1Qbv-2015 (Amendment to IEEE Std 802.1Q-2014 as amended by IEEE Std 802.1Qca-2015, IEEE Std 802.1Qcd-2015, and IEEE Std 802.1Q-2014/Cor 1-2015) , vol., no., pp.1-57, 18 March 2016, doi: 10.1109/IEEESTD.2016.8613095.

levy:
Ah, I see. I misunderstood what kind of limit you are talking about. Yes,
this makes sense, although it's not trivial to implement.

yue2388253:
Currently the time aware shaper is implemented via PeriodicGate, which is configured individually, and is different from the IEEE 802.1Qbv standard where gates are configured by a GCL.
Is there any plan to re-implement it? 
I am currently doing some research on the GCL synthesis problem, and maybe I can be of soem help.

levy:
[...]

PR #774, opened 2022-06-27
Ieee8021qTimeAwareShaper: added GateControlList

yue2388253:
Ieee8021qTimeAwareShaper: added GateControlList, a module which accpets parameters in GCL format and configures the corresponding PeriodicGate modules.

### DIFF
--- a/showcases/tsn/trafficshaping/timeawareshaper/omnetpp.ini
+++ b/showcases/tsn/trafficshaping/timeawareshaper/omnetpp.ini
@@ -43,7 +43,13 @@ description = "Traffic shaping using time-aware shapers"
 *.switch.eth[*].macLayer.queue.numTrafficClasses = 2
 *.switch.eth[*].macLayer.queue.*[0].display-name = "best effort"
 *.switch.eth[*].macLayer.queue.*[1].display-name = "video"
-*.switch.eth[*].macLayer.queue.transmissionGate[0].offset = 0ms
-*.switch.eth[*].macLayer.queue.transmissionGate[0].durations = [4ms, 6ms] # period is 10 # length of periods
-*.switch.eth[*].macLayer.queue.transmissionGate[1].offset = 6ms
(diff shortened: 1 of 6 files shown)


yue2388253:
This PR resolved #771

yue2388253:
This implementation may lead to unexpected guard bands. For example, assume we have the following configuration

```ini
*.s0.eth[*].macLayer.queue.gateControlList.offset = 0ms 
*.s0.eth[*].macLayer.queue.gateControlList.durations = [4ms, 2ms, 2ms, 2ms]
*.s0.eth[*].macLayer.queue.gateControlList.gateStates = ["01", "00", "10", "00"]
```

which would generate the following configuration for each PeriodicGate

```ini
*.s0.eth[*].macLayer.queue.transmissionGate[0].offset = 0ms
*.s0.eth[*].macLayer.queue.transmissionGate[0].durations = [4ms, 6ms] # period is 10 # length of periods
*.s0.eth[*].macLayer.queue.transmissionGate[1].offset = 0ms
*.s0.eth[*].macLayer.queue.transmissionGate[1].durations = [6ms, 2ms, 2ms, 0ms]
```

For `transmissionGate[1]`, a `0ms` duration is appended at the end of the list 
to satisfy the requirement---the number of the list should be even.
This would lead to extra `changeTimer` events at PeriodicGate.
Since PeriodicGate would not allow a packet to pass through if the
the complete transmission of the packet cannot finish before the `changeTimer`
event, this can lead to unexpected guard bands.


```cpp
clocktime_t flowEndTime = getClockTime() + s((packet->getDataLength() + extraLength) / bitrate).get() + SIMTIME_AS_CLOCKTIME(extraDuration);
return !changeTimer->isScheduled() || flowEndTime <= getArrivalClockTime(changeTimer);
```

levy:
Maybe the solution for the guard band issue is not to introduce a parameter to avoid the guard bands before closing and immediately re-opening the gate. I think the module should make it possible to use it without any zero simulation time gate closing/opening with any circular gate close/open durations.

PR #776, opened 2022-06-29
[TSN feature] Ieee8021qTimeAwareShaper and PeriodicGate

yue2388253:
* Ieee8021qTimeAwareShaper: added GateControlList, a module which accepts parameters in GCL format and configures the corresponding PeriodicGate modules.
* PeriodicGate: provide a parameter to configure whether the implicit guard band feature is enabled.

### DIFF
--- a/showcases/tsn/combiningfeatures/invehicle/omnetpp.ini
+++ b/showcases/tsn/combiningfeatures/invehicle/omnetpp.ini
@@ -422,6 +422,9 @@ extends = TimeSensitiveNetworkingBase
 *.gateScheduleConfigurator.typename = "AlwaysOpenGateScheduleConfigurator"
 *.gateScheduleConfigurator.gateCycleDuration = 500us
 
+# disable implicit guard bands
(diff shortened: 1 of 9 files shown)


levy:
Thanks for the PR and sorry for the delay, I was on holiday.

I think the additional parameter in the PeriodicGate module can be merged as it is.

As for the GateControlList submodule I'm not sure. It is generally correct, but I think this submodule should be optional, becuase the gate submodule types can be changed to anything. Also, in the case of automatic gate schedule configuration, the GateControlList submodule is not used and thus it would be confusing if not wrong to have it. I'm not sure if the GateControlList submodule should be used by default in the example. I kinda understand the reason is that you wanted to make the example more standard compliant, so I can be convinced about this.

levy:
This pull request has been merged into master with some changes:
 - examples/showcases are not modified
 - the GateControlList module is not present in the Ieee8021qTimeAwareShaper by default, it must be requested using hasGcl

PR #804, opened 2022-07-22
Inet 3.8.3 update 2 files

NikolaosGian:
ScenaOsgEarthVisualizer.h/cc


NikolaosGian:
Those changes will help every user to compile the inet file right at first time with omnet++ 6.0.
This happend to me when i try to build the inet file for the first time without those changes

rhornig:
Thanks a lot for the PR. 

A tip for the future: you can use a single commit for changes that are logically belonging together and for the commit message, you should state what was the *intention* of the changes. The actual change is visible in the diff anyway, but it is useful to see *why* you had a motivation for the change.

PR #821, opened 2022-10-13
ScenarioManager: Add fault injection module for use in MANET simulations

n-jay:
Node faults in MANET implementations can currently be [simulated](https://inet.omnetpp.org/docs/users-guide/ch-lifecycle.html) using `ScenarioManager`.
However, this [uses](https://inet.omnetpp.org/docs/users-guide/ch-scenario-scripting.html) the `ScenarioScripting` system which has several limitations 

- Pre-scripted node failures unrepresentative of real world
- Static failure events do not scale with node numbers when conducting parametric studies

This PR introduces a [Chaos Engineering](https://principlesofchaos.org/) module to OMNET simulations that can inject lifecycle events to random nodes during the simulation. 
Currently, it supports hard fault types such as **crashes** and **shutdowns**.

### DIFF
--- a/src/inet/common/chaos/ChaosEvent.msg
+++ b/src/inet/common/chaos/ChaosEvent.msg
@@ -1,5 +1,3 @@
-import inet.common.INETDefs;   
-
 namespace inet;
 
 message ChaosEvent 
--- a/src/inet/common/chaos/ChaosEvent.msg
+++ b/src/inet/common/chaos/ChaosEvent.msg
@@ -1,0 +1,9 @@
+import inet.common.INETDefs;   
+
+namespace inet;
+
+message ChaosEvent 
+{
+    int command;
+    int nodeNum;
+}
--- a/src/inet/common/chaos/ChaosManager.cc
+++ b/src/inet/common/chaos/ChaosManager.cc
@@ -21,7 +21,6 @@
 #include <string>
 #include <cstring>
 #include <map>
-//#include "inet/routing/leach/Leach.h"
 #include "inet/common/lifecycle/LifecycleOperation.h"
 #include "inet/common/lifecycle/ModuleOperations.h"
 #include "inet/common/chaos/ChaosEvent_m.h"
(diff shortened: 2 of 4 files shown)


levy:
I'm reluctant to merge something like this for two reasons:
 - the module seems to generate a very specific chaotic node crash/shutdown behavior
 - the next OMNeT++ version will have full Python support for all kinds of scripting dynamic behavior into the simulation

In fact, the second reason would make the current ScenarioManager pretty much obsolete.
You will be able to simply write a python file, where you can schedule events which crash/shutdown/startup network nodes in a few lines of code.

n-jay:
Hi @levy 
> the module seems to generate a very specific chaotic node crash/shutdown behavior

Yes, in its current phase I'm only modeling randomized node shutdowns and crashes. I was hoping to accomplish that in the first phase and move on to modeling chaotic events of other parameters after making the necessary changes based on the code review (such as adding unit tests etc.).
In the end, I'm planning to have the same coverage of parameters as ScenarioManager.

> the next OMNeT++ version will have full Python support for all kinds of scripting dynamic behavior into the simulation

> You will be able to simply write a python file, where you can schedule events which crash/shutdown/startup network nodes in a few lines of code.

My main issue even with the XML approach in ScenarioManager was that it _was scripted_. I wanted to approach this as a [Chaos Engineering](https://principlesofchaos.org/) module where faults occurred randomly rather than as a pre-scripted event generator.
I'm more than happy to make any necessary changes the INET team recommends so that it furthers this goal.

> In fact, the second reason would make the current ScenarioManager pretty much obsolete.

My module has no dependencies with ScenarioManager and it was a mistake on my part renaming it as such.
This is for all intents an purposes, a completely standalone module.

n-jay:
Hi @levy 
would appreciate a followup on this.

levy:
Thank you for your contribution, we appreciate your work but we don't plan to merge it.

As I have mentioned above, even the current ScenarioManager is now considered obsolete by the upcoming OMNeT++ Python integration. The Python solution is much more flexible, concise, easier to combine, and it is still efficient enough to be used in simulations.

Please check the presentations on 2022 OMNeT++ Summit webpage what is coming.

Issue #848, opened 2023-01-05
Credit accumulation on closed transmission gates


Clunphumb:
Hello,
I am trying to combine Credit-Based Shaping with Time-Aware Shaping and stumbled upon a problem that is hindering me in simulating my network.
I noticed that queues, which use the Ieee8021qCreditBasedShaper as transmission selection algorithm, are accumulating credit while their respective transmission gate is closed. This leads to an increase of burstiness in queues as they are able to send much more of their data right after reserved time-slots. Moreover this leads to a shift of the earliest point in time a lower prioritized queue is able to transmit and therefore an increase in delay for this queue.
An alternative to the current behavior would be to freeze the current credit of queues whose transmission gate is currently closed as proposed in a note in the Standard IEEE Std 802.1Q-2018 p.201.
However, I don't know how to approach this problem myself and would be grateful for some advice on the issue.

levy:
You should look at the CreditBasedGate module for the implementation of the credit-based shaping algorithm. You could change this module to check if a packet can be pulled from the connected module on its input. If no packet is available then the number credits should not increase.

Clunphumb:
Hi levy,
thank you for your reply. Sadly I wasn't able to implement this solution as I'm not yet familiar with the framework and I didn't completely understand your solution. Am I supposed to check if I can pull from CreditBasedGate or from the successor of the CreditBasedGate? If the ladder is the case i'm unsure on how to get the successor of the successor of the CreditBasedGate to check that. Could you point me in the right direction?
On another note, I found that PacketGateBase defines a variable called `isOpen_` which for me seems to be something I could maybe use somehow? 

Thanks in advance

Clunphumb:
For anyone interested in this: I think I managed to implement it the way I want it to work by adding a `ModuleRef<PeriodicGate>` to the CreditBasedGate.h and initialized with `periodicGate.reference(outputGate, false);`. I also added the Method `virtual bool isTransmissionGateOpen() const {return periodicGate->isOpen();}` and used it in `updateCurrentCreditGainRate()` to set `currentCreditGainRate = 0`. I don't know much about C++ or OMNeT or INET but this approach seems to be working.

PhilippComet:
Hi, unfortunately i have the same issue. 
@Clunphumb Thank you for the approach, but can you please explain a bit more detailed your modifications? Where did you add `periodicGate.reference(outputGate, false);`? 

Additionally here a visualization of the problem of credit accumulation during a closed TAS Gate

[...]

PR #852, opened 2023-01-24
CreditBasedGate: Credit updates on closed gate

Clunphumb:
This change addresses the issue that the credit of CreditBasedGates was increasing on  while the corresponding PeriodicGate was closed. Per IEEE 802.Q-2018 credit should be frozen on closed gates. Moreover an option was added to toggle the accumulation of credit during pre-closing time, i.e. when the PeriodicGate is open but not long enough anymore to send the next package. According to IEEE credit should still increase, which leads to higher usage of bandwidth but fairness between different queues isn't as good anymore and credit could be unbounded. Therefore a parameter allows to switch the behavior to accommodate different use cases. For reference see #848

### DIFF
--- a/src/inet/queueing/gate/CreditBasedGate.cc
+++ b/src/inet/queueing/gate/CreditBasedGate.cc
@@ -33,14 +33,17 @@ void CreditBasedGate::initialize(int stage)
         maxCredit = par("maxCredit");
         displayStringTextFormat = par("displayStringTextFormat");
         currentCredit = par("initialCredit");
+        isAccumulatingBeforeGateClose = par("isAccumulatingBeforeGateClose");
         currentCreditGainRate = idleCreditGainRate;
         lastCurrentCreditEmitted = currentCredit;
         lastCurrentCreditEmittedTime = simTime();
         isOpen_ = currentCredit >= transmitCreditLimit;
+        periodicGate.reference(outputGate, false);
         cModule *module = getContainingNicModule(this);
         module->subscribe(transmissionStartedSignal, this);
         module->subscribe(transmissionEndedSignal, this);
         module->subscribe(interpacketGapEndedSignal, this);
+        module->subscribe(gateStateChangedSignal, this);
         changeTimer = new cMessage("ChangeTimer");
         WATCH(currentCredit);
         WATCH(currentCreditGainRate);
@@ -50,6 +53,7 @@ void CreditBasedGate::initialize(int stage)
         updateCurrentCreditGainRate();
         emitCurrentCredit();
         scheduleChangeTimer();
+        isInPreClosingTime();
     }
     else if (stage == INITSTAGE_LAST)
         updateDisplayString();
@@ -68,6 +72,8 @@ void CreditBasedGate::handleMessage(cMessage *message)
         updateCurrentCreditGainRate();
         // 5. reschedule change timer based on currentCredit and currentCreditGainRate
         scheduleChangeTimer();
+        // 6. check if pre-closing time has begun
+        isInPreClosingTime();
     }
     else
         throw cRuntimeError("Unknown message");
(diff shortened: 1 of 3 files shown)


levy:
Thanks for the submission! Could you please merge the commits, it would be nice to have this PR as a single patch.

levy:
It looks like that you pushed a merge commit into the PR. You should force push your branch with the single commit that contains all changes instead. Ideally, this PR should contain a single patch relative to INET master.

Clunphumb:
I hope I managed to do it correctly now?

levy:
Looks better now, we still have to review it though. :-)

levy:
Now that the INET 4.5 release date is approaching, I took a deeper look at this PR.

I have some comments. For example, I noticed that the PR assumes that the output gate of a CreditBasedGate module is connected to a PeriodicGate module. The PR allows the periodicGate module reference to be nullptr, if there is no PeriodicGate connected, but the code will crash when the reference is actually used in such a case.

Also, the canPacketBeTransmitted method calls canPullSomePacket on the connected periodicGate module with the wrong cGate parameter. The correct cGate in that call would be the output gate of the periodicGate.

Another issue is that the method name isInPreClosingTime is a bit confusing, because the gate is not just simply before being closed when it returns true. This name should imply that the next packet cannot be pulled/pushed through the connected periodic gate module before the gate will close. I don't have a better name yet.

I'm not sure that the current approach of subscribing to the gateStateChanged signal on the network interface is right. Well, the subscription could be done on the periodicGate itself, but I'm not even sure if that is what I would do. There are other ways to communicate with the connected module, namely the handleCanPushPacketChanged call that comes from the PeriodicGate when it opens or closes.

Another issue that I have is if I create a different implementation that gives the same behavior, I cannot test the new implementation, because there's no example which I can use to test it.

I would like to include this behavior change in the next INET release, but I'm not sure how to do that with the current state of this PR. I will think a bit more about this. Please comment!

levy:
It would be nice to get an example simulation (NED + INI files) where the incorrect behavior can be seen, and where the applied PR fixes this behavior. Thanks!

levy:
I have extracted and somewhat modified the first part of the PR that handles the case when the transmission gate is closed. Could you please take a look at cebc04c23129858817378cbe874d442cfb51f21d and tell me if this looks good to you?

levy:
The first part has been merged into INET master. I have also added a new showcase under tsn/trafficshaping/cbsandtas, it still needs more work to make this a meaningful example of combining CBS with TAS, though.

levy:
I'm considering adding a new signal to PeriodicGate that would mark the start/end of each guard band interval. This would greatly simplify the implementation of CreditBasedGate with respect to not increasing the number of credits during the implicit guard band. What do you think?

Unfortunately, I still don't have any of your examples where I could actually test this.

Clunphumb:
Thank you very much for your feedback and your work on this PR. 

I already tested and checked the credit accumulation freeze on the closed gate which seems to be working as expected. However, I am unsure if my implementation of stopping credit accumulation during the pre-closing time is correct now. I will try to test this again as soon as I can. I will provide an example configuration tomorrow.

A new signal in PeriodicGate seems to be a very good idea as far as I can tell. At the very least this would declutter my current implementation.

Clunphumb:
I attached the corresponding files for testing, however I am not yet satisfied with the behavior if credit is not supposed to increase during pre-closing time. I noticed some mistakes in the calculations and will have to fix them first.

[omnetpp.ini.txt](https://github.com/inet-framework/inet/files/10978768/omnetpp.ini.txt)
[CreditBasedGate.ned.txt](https://github.com/inet-framework/inet/files/10978779/CreditBasedGate.ned.txt)

Former incorrect behavior
former_behavior

Correct stop of accumulation during closed gates:
accumulation_in_pre_closing_time

levy:
Thanks for the reply! So, I created a new branch called 'topic/cbsandtas' that adds the guard band state changed signal to PeriodicGate and also changes the CreditBasedGate to stop accumulating credits during the implicit guard band of the connected PeriodicGate. I checked the statistics and the results seem to be ok. See the following chart for example:

image

Could you please verify if this branch satisfies your needs?

levy:
With the latest changes related to #848 all remaining parts of this PR are in INET master in one form or another.

Issue #854, opened 2023-01-30
Src mac address of the arp package during emulation is 00:00:00:00:00:00


Jimmy01240397:
I'm trying to edit [emulation/pinger](/inet-framework/inet/tree/master/examples/emulation/pinger) and make it can ping another host on LAN.
But when I use tcpdump to capture traffic I saw the src mac address of the arp package is 00:00:00:00:00:00 and arp didn't response then ping failed. Why?

![image](https://user-images.githubusercontent.com/57281249/215602139-60ea1c46-3c86-42b5-83f2-c926894f3e9f.png)

ZoltanBojthe:
For reproduce your bug, need more information:
Which inet version was used when the error occurred?
Could you attach your modified ned/ini files?

Your patch is unnecessary, the EthernetMacHeaderInserter and EthernetEncapsulation modules contain code for replace 00-00-00-00-00-00 source MacAddress to MacAddress of the sender interface.

ZoltanBojthe:
Fixed in INET master.

PR #856, opened 2023-02-13
Fix missing source mac address of the packet

Jimmy01240397:
Fix the problem of issue #854


levy:
The issue #854 related to this PR is already fixed in Arp differently. The source MAC address is filled in the MacAddressReq by the ARP protocol module with the same value that is also added to the ARP packet. The IP protocols don't set the source MAC address in the request intentionally, because they only decide the outgoing interface. The link layer protocols fill in the source MAC address automatically with the network interface MAC address if the src address was unspecified in the Req. The EthernetEncapsulation module has been moved from the network interface to the link layer of the network node, this caused the above src MAC address filling process to fail, but this issue has also been fixed meanwhile.

Issue #860, opened 2023-02-23
[IEEE 802.1Qci PSFP] StreamFilter does not check StreamInd Tag of frames and drops them


olfline:
Hello dears,

I have a problem with the StreamFilter in the Ieee8021qFilter and SimpleIeee8021qFilter Module.
The StreamFilter drops frames of a stream which should be forwarded.

### Problem:
I have a simple Network as follows ([ned.txt](https://github.com/inet-framework/inet/files/10816670/ned.txt)): 
Client <---> Switch <---> Server
When the stream-frames are facing the StreamFilter in the Switch, they get dropped, even though the stream should match.

###  Cause:
The use of the StreamFilter Module inside the (Simple)Ieee8021qFilter compound Module causes the frames to have the _StreamInd_ Tag instead of the _StreamReq_ Tag. I don't know why. SimpleIeee8021qFilter and Ieee8021qFilter behave the same.
I saw in the src that the StreamFilter only checks for the _StreamReq_ label and not the _StreamInd_ label. Therefor all frames are dropped - even if the stream should match. The picture shows the StreamInd Tag of an example frame.
![StreamIndTag](https://user-images.githubusercontent.com/50140648/220998125-df836272-b4ea-4826-9620-2e74cc5dc657.png)

###  Possible solution:
Add check for the _StreamInd_ Tag in StreamFilter Module.

###  Note:
Why does the StreamFilter change the Tag in the frame?

Omnetpp  Version: 6.0.1, Build id: 220831-e4f07d04b8
Inet commit [d27179e](https://github.com/inet-framework/inet/commit/d27179eb41e5e74109a47e7efa622cdad5b510c2)

Can you please verify?
Thank you very much!

levy:
The StreamInd and StreamReq tags mean different things. StreamInd specifies the stream that the incoming packet is part of. In contrast, StreamReq specifies the stream that the outgoing packet is part of. Most often StreamInd is turned into a StreamReq when the packet gets forwarded in a switch. This happens because the outgoing stream is usually the same as the incoming one. Please also note that if there is no StreamReq on an outgoing packet, then the packet is not going to be part of any stream, and it will not be encoded using 802.1 Q-TAG and 802.1 R-TAG on the wire.

With all that said, it still makes sense to filter for StreamInd tag, namely in the ingress filter part of the bridging layer. The decision between filtering for StreamReq or StreamInd should not be made automatically, but it should be requested by the user through NED parameters.

levy:
[...]

PR #861, opened 2023-02-24
StreamFilter looks now at StreamInd Tag of packets

olfline:
This is addressing the issue [860](https://github.com/inet-framework/inet/issues/860) that the StreamFilter does not look at the StreamInd Tag of packets.


levy:
Thanks for the PR! The related issue has been fixed in the master branch with a somewhat different approach.

PR #879, opened 2023-04-05
Revert "ProtocolGroup: deallocate dynamically added protocols in the destructor"

aagon:
This commit assumes that every `Protocol` added with `ProtocolGroup::addProtocol` has been dynamically allocated (it admits as much in the comment).

Firstly, this assumption is simply wrong. It is perfectly possible to have a `Protocol` as a global variable. In this case, the destructor of the `ProtocolGroup` class attempts to free an address of the data segment, which produces an error at the end of a simulation ("free() : invalid pointer").

Secondly, this commit also assumes (implicitly this time) that a dynamically allocated `Protocol` hasn't already been freed by the time the destructor of `ProtocolGroup` is called, which turns out to be also wrong. If we transform our previous `Protocol` instantiated as a global variable in a dynamically allocated one, whose pointer is stored in a global `Protocol *` variable, then we get another error at the end of a simulation (more along the lines of "free() : double free or corruption").

I am not entirely sure why this commit was added in the first place. It seems to be inside a series of commits that change the `ProtocolGroup` API, but is AFAICT independent from it.

If someone has dynamically allocated a `Protocol`, it should be their own responsibility to free it, and it shouldn't have anything to do with the fact that they have decided to register said `Protocol` inside a `ProtocolGroup`.

Please revert this commit.

Best regards,

Aymeric Agon-Rambosson

### DIFF
--- a/src/inet/common/ProtocolGroup.cc
+++ b/src/inet/common/ProtocolGroup.cc
@@ -22,12 +22,6 @@ ProtocolGroup::ProtocolGroup(const char *name, const Protocols& protocolNumberTo
     }
 }
 
-ProtocolGroup::~ProtocolGroup()
-{
-    for (auto p : dynamicallyAddedProtocols)
-        delete p;
-}
-
 const Protocol *ProtocolGroup::findProtocol(int protocolNumber) const
 {
     auto it = protocolNumberToProtocol.find(protocolNumber);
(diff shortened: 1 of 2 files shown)


levy:
I think you are right. Besides the problems you have already mentioned, the current code crashes if the same dynamically allocated protocol is added to multiple protocol groups, because each protocol group will delete the added protocol.

haug-den-lucas:
Hi, I realized this fix is still not part of any release version of INET yet, even though the fix is in master for almost 2 years now.
Is there any plan to include this fix in the next releases?

levy:
Yes of course, this change will be included in the upcoming INET 4.6 version, until then you can use the master branch.

PR #910, opened 2023-09-18
Improve code readability for receiving signals in `Gptp.cc`

delvh:
Previously, this method was absolutely unreadable.
Now, it is more readable.
The logic should not have changed.


avarga:
Merged, thanks for your contribution!

PR #924, opened 2023-11-08
Fix merging of two ByteCountChunks with different data

Musteblume:
Just increasing the length of a ByteCountChunk when inserting a ByteCountChunk with different data leads to incorrect packets.

Example: Ethernet padding (PaddingInserter) with 0 when payload is also a ByteCountChunk with default 0x3f data.


levy:
You are right, thanks for the PR!

Issue #928, opened 2023-12-06
LEACH routing protocol


n-jay:
Including [LEACH](https://ieeexplore.ieee.org/abstract/document/926982) would be a good addition to the current protocol selection considering it is one of the more widely used seminal clustering protocols.

levy:
I close this issue because there's an open PR for the same topic.

PR #929, opened 2023-12-06
Add LEACH routing protocol

n-jay:
Adds the [LEACH](https://ieeexplore.ieee.org/abstract/document/926982) protocol to existing routing protocol selection.

Addresses issue https://github.com/inet-framework/inet/issues/928


n-jay:
Fingerprint values seem to be changing with each subsequent run.

rhornig:
1st. Thanks a loft for the PR. We are busy with other topics right now, but we will review this for the next INET release. Until then:

As other fingerprints are stable, this indicates that the code execution is not repeatable. In my experience this might be caused by:

- some uninitalized variables lurking in the code (not very likely as it would rather crash the process)
- using undefined behavior in C++ code (not likely: hopefully you don't do any nasty things in your code :)
- iterating over an unordered collection (this is the most likely cause)

It is very easy to create a hashed map or set and forget about the fact that its iteration order may depend on memory layout. We had some similar issues in early versions of INET and OMNeT++.

rhornig:
I just had a quick look into the code (not a proper review, but some comments).

- license is fine (LGPL), but please use similar format for file headers like the rest of the code (i.e. use just SPDX-License-Identifier: LGPL-3.0-or-later) This reduces noise and still does the job.
- including headers files must not assume that the current directory is on the include path (i.e. prefix with inet/routing/leach even for your local headers). This is done because some other projects using INET may have a local header file similarly named (for whatever reason) and hell will broke...
- I would put the whole codebase under the inet::leach namespace (instead of the top level ::inet)
- use proper enums. `packetType == 5` does not mean anything to the code reader. It's an implementation detail that the value is currently 5.
- the check for a single wireless interface is very fragile. You should not assume that a wireless interface is called "wlan". It may change in the future, or you may want to use leach on a mobile connection (i.e. from Simu5G). I would rather have a parameter on the protocol module that species the interface that must be used. Its default value can be "wlan0", so you don't have to count the interfaces or know the name in advace). If need arise the user still can change the default value.
- finally regarding the changing fingerprints. I see an include <unordered_set> in LeachBS.cc. Check that file carefully whether the does not depend of that set's order.

n-jay:
Noted @rhornig, thank you!
I'm still very much a novice when it comes to C++ so this was extremely helpful. I'll look into the points you made and do the necessary changes.

Can those changes be committed separately beyond this point (with detailed commit messages), or should I squash them to the earlier commits?

rhornig:
just go on with new commits. Once we decide to merge, we will squash it.

n-jay:
> I see an include <unordered_set> in LeachBS.cc. Check that file carefully whether the does not depend of that set's order.

@rhornig I looked into the use of unordered collections in the `LeachBS.cc` file and it turned out to be an unused `#includes` from my original code which I've since removed in https://github.com/inet-framework/inet/pull/929/commits/298eff7e475b8d35a7ab857ff7b55f957e9b165d. 
I've gone over the rest of my `.cc` and `.h` files and verified that there aren't any more unordered collections, only vectors. However the problem still seems to persist.

rhornig:
Regarding the failing fingerprint tests: If there is no obviously visible parts where unordered collections are used then I would check also for memory issues. Running valgrind memcheck may provide some tips. I would also check whether release and debug versions produce the same fingerprint. If they do, I would try to somehow reproduce the different fingerprints reliably locally  (having a different fingerprint only during the github ci test makes catching this bug very hard).

Once you can reproduce the fingerprints, you can write out an event log file for both runs and compare those and see where the first difference happens. That would at least give some hints what is happening.

These are the ugliest type of errors one can find in a simulation. That's the main reason we have implemented fingerprinting, to at least catch them early.

rhornig:
We will need some documentation about the routing protocol implementation in the Leach.ned file.:
- What protocol it implements (possibly an URL to the standard)
- what features are / are not implemented compared to that standard document. WHat is the difference compared to the standard.
- implementation specific documentation, like what parameters must be configured by the user and what are th elimitation (e.g. does it work only on wireless interfaces? Is there limitations on what kind of wireless protocol can be used? Wifi, 5G etc. Are there limitation on the number of interfaces (as far as I see only a single wireless interfaces is supported).

rhornig:
For the interface name parameter: We have similar code patterns (i.e. for selecting the name of the interface to be used) in Dymo and GPSR. The parameter is called `interfaces` and you can use a `cPatternMatcher` to iterate and match the selected interface (see the related .cc files). This approach gives the advantage that you can use pattern characters in the name of the interface and potentially also allows to support binding to more than one interface. If your implementation support only a single interface, I would still name it `interfaces` and would comment there that only a single interface is allowed (as it is an implementation restriction). You can still throw errors if more than one interface name matches.

https://github.com/inet-framework/inet/blob/master/src/inet/routing/dymo/Dymo.cc#L1294

n-jay:
> Regarding the failing fingerprint tests: If there is no obviously visible parts where unordered collections are used then I would check also for memory issues. Running valgrind memcheck may provide some tips. I would also check whether release and debug versions produce the same fingerprint. If they do, I would try to somehow reproduce the different fingerprints reliably locally (having a different fingerprint only during the github ci test makes catching this bug very hard).

Will look into this @rhornig 
I also have a hunch on what might be causing this. So will check it out, run the simulation and get back to you.

> We will need some documentation about the routing protocol implementation in the Leach.ned file.:

Noted, will add that. 
Would documentation that looks like the [Dymo.ned](https://github.com/inet-framework/inet/blob/master/src/inet/routing/dymo/Dymo.ned) and [Gpsr.ned](https://github.com/inet-framework/inet/blob/master/src/inet/routing/gpsr/Gpsr.ned) be sufficient?

> For the interface name parameter: We have similar code patterns (i.e. for selecting the name of the interface to be used) in Dymo and GPSR. 

Noted, will study them and see how the code can be modified.
My initial code was based on the Dsdv implementation which I think is a bit dated.
https://github.com/inet-framework/inet/blob/3a440549217a5eb84d259e1e99fbeb46742edb7e/src/inet/routing/dsdv/Dsdv.cc#L72-L83

Once this is merged, I can work on updating that to match Dymo and Gpsr if its ok.

rhornig:
Documentation: keep it short. Refer to the original specification and mention only differences / limitations compared to that and parameters and behavior specific to this current implementation (i.e. mention possible statistics gathered, limitation (like only a single wireless interface supported etc). Think about a NEW user who is otherwise already familiar with the leach protocol (by reading the spec).

n-jay:
Hi @rhornig 
Thanks for suggesting Valgrind Memcheck, ran it via the Eclipse plugin and turns out the fingerprint anomaly was due to an uninitialized variable which I've fixed in https://github.com/inet-framework/inet/pull/929/commits/57c36ed7279cd4217baafa52ec743ee7f37fe56f. 

I also updated the network interface configuration in https://github.com/inet-framework/inet/pull/929/commits/9c05df0704f0f4c751ed6b97ca43bb8f2cf055b2. It seems the current DSDV implementation has been slightly tweaked from the older version I used in INET 4.2.5, so took inspiration from that + the code pattern you mentioned in https://github.com/inet-framework/inet/pull/929#issuecomment-1880960597.

The docs are also included in the NED file in https://github.com/inet-framework/inet/pull/929/commits/6947423ec1ba8f7c635624dcdd005f9bd0978bc7. Hope this is sufficient.

n-jay:
Hi @rhornig, kind follow-up on this.

src/inet/routing/leach/Leach.ned
```diff
@@ -0,0 +1,38 @@
...
+        @class("leach::Leach");
+        @display("i=block/routing");
+        
+        // context parameters
+        string interfaceTableModule; 
+        
+        // LEACH parameter group
+        double maxVariance = default(1);
+        int numNodes = default(10);
+        double clusterHeadPercentage = default(0.5);
```
rhornig, line 34: Missing documentation and non descriptive / misleading name. Generally, if you specify a default value in a NED file, then that value must be supported by the specification document somewhere. If this is just a random picked value, then do NOT set it as a default value, but force the user to set it explicitly in the INI file. Besides the name is suspicious. You meant half a percent as a default value here? or rather it should be named `clusterHeadRatio` and then the value represents 50% ?

recommendation: `double clusterHeadRatio @unit(ratio) = default(0.5ratio)` and in that case, you can even use = 50pct because omnet can convert automatically between ratio and pct

src/inet/routing/leach/Leach.cc
```diff
@@ -0,0 +1,450 @@
...
+
+        clusterHeadPercentage = par("clusterHeadPercentage");
+        numNodes = par("numNodes");
+
+        roundDuration = dblrand(0) * 10;
+        TDMADelayCounter = 1;
+        event = new cMessage("event");
+        roundDuration = dblrand(0) * 10;
+        wasCH = false;
+
```
rhornig, line 50: The following vectors are rather suspicious to me. you are reading the numNodes parameter, then you create some vectors with this size and then throw away those vectors. Note that these are local variables on the stack that has nothing to do to the similarly named member variables in the class. You are not initialize those variables here, you are just hiding them. As you are dynamically fill those vectors, you don't have to pre-allocate those vectors anyway. Just delete these lines and the `numNodes` parameter.

src/inet/routing/leach/Leach.cc
```diff
@@ -0,0 +1,450 @@
...
+void Leach::stop() {
+    nodeMemory.clear();
+    nodeCHMemory.clear();
+    extractedTDMASchedule.clear();
+    TDMADelayCounter = 1;
+}
+
+void Leach::configureInterfaces() {
+    int numInterfaces = interfaceTable->getNumInterfaces();
```
rhornig, line 79: This code picks the first wireless interface from the interface table. This must be documented in the NED file (i.e. that only the first can be used). I would rather go with pattern we use in other similar cases where you have an `interfaces` or `interface` string parameter and the user MUST explicitly specify which interface should be used by the module (you can use the cPatterMatcher similarly to the other MANET routing protocols)

Let's assume there is a host which has both WiFi and a 802.15.4 radio and WiFi happens to be on wlan0 and the sensor is on wlan1. with this code it is not possible to conigure wlan1 for Leach because wlan0 would be picked.

src/inet/routing/leach/Leach.cc
```diff
@@ -0,0 +1,450 @@
...
+}
+
+void Leach::handleMessageWhenDown(cMessage *msg) {
+    delete msg;
+    OperationalBase::handleMessageWhenDown(msg);
+}
+
+void Leach::handleSelfMessage(cMessage *msg) {
+    if (msg == event) {
+        if (event->getKind() == SELF) {
```
rhornig, line 131: I don't see the SELF kind necessary. you already know that this is a self message as `msg == event`. Also there is an explicit `isSelfMessage()` on the messages which gives exactly this info if needed.

src/inet/routing/leach/Leach.h
```diff
@@ -0,0 +1,125 @@
...
+        Ipv4Address nodeAddr;
+        Ipv4Address CHAddr;
+        double energy;
+    };
+
+    struct TDMAScheduleEntry {
+        Ipv4Address nodeAddress;
+        double TDMAdelay;
+    };
+
```
rhornig, line 62: Are you sure vector<> is the best data structure for this? As far as I see, you are often looking up using an ip address and iterate over the whole vector. This implies an o(2) complexity with node number. A hash-map would be probably a better approach (I'm not sure on this)

src/inet/routing/leach/LeachBS.cc
```diff
@@ -0,0 +1,103 @@
...
+
+}
+
+void LeachBS::stop() {
+
+}
+
+void LeachBS::configureInterfaces() {
+    int numInterfaces = interfaceTable->getNumInterfaces();
+    for (int i = 0; i < numInterfaces; i++) {
```
rhornig, line 58: This just pick the fors wireless interface. an explicit `interface` parameter would be better with wlan0 as a default.

rhornig (changes requested):
Did a very quick review again with in-line recommendations.

examples/manetrouting/leach/Leach.ned
```diff
@@ -0,0 +1,37 @@
...
+package inet.examples.manetrouting.leach;
+
+import inet.networklayer.configurator.ipv4.Ipv4NetworkConfigurator;
+import inet.node.contract.INetworkNode;
+import inet.node.inet.SensorNode;
+import inet.physicallayer.wireless.common.contract.packetlevel.IRadioMedium;
+import inet.visualizer.contract.IIntegratedVisualizer;
+import inet.environment.common.PhysicalEnvironment;
+
+network Leach
```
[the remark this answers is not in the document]
n-jay: Resolved in https://github.com/inet-framework/inet/pull/929/commits/e2c5cb03c51b9c441979e453cac368908641c41f

examples/manetrouting/leach/Leach.ned
```diff
@@ -0,0 +1,37 @@
...
+{
+    parameters:
+        int numNodes;
+        @display("bgb=200.8119,200.0099;bgg=100,1,grey95");
```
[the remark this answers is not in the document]
n-jay: Resolved in https://github.com/inet-framework/inet/pull/929/commits/9ed810b66b30bff4638d314b85f70163e2ad4a38

examples/manetrouting/leach/omnetpp.ini
```diff
@@ -0,0 +1,86 @@
+[General]
+network = Leach
+
+num-rngs = 2
+rng-class = omnetpp::cMersenneTwister
```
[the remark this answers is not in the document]
n-jay: Resolved in https://github.com/inet-framework/inet/pull/929/commits/a6aeb1d8297ed3acea88aca2677d9b771a1d3a32

src/inet/node/leach/LeachBS.ned
```diff
@@ -0,0 +1,19 @@
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+//
+
+package inet.node.leach;
+
+import inet.node.inet.AdhocHost;
+import inet.routing.leach.LeachBS;
+
+module LeachBS extends AdhocHost {
```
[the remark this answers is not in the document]
n-jay: Sorry about that, missed those NED files. Resolved in https://github.com/inet-framework/inet/pull/929/commits/85481e9b7706316ebeb729a40a257f3f6fbad2c2

src/inet/node/leach/LeachNode.ned
```diff
@@ -0,0 +1,19 @@
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+//
+
+package inet.node.leach;
+
+import inet.node.inet.AdhocHost;
+import inet.routing.leach.Leach;
+
```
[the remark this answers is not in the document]
n-jay: Resolved in https://github.com/inet-framework/inet/pull/929/commits/85481e9b7706316ebeb729a40a257f3f6fbad2c2

src/inet/node/leach/LeachNode.ned
```diff
@@ -0,0 +1,19 @@
...
+module LeachNode extends AdhocHost {
```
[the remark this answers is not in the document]
n-jay: The standard LEACH protocol doesn't specify the protocol for the Physical and MAC layers, however for my implementation I've used 802.15.4 standard that covers those layers as I was simulating a WSN as realistically as possible. Hence the reference to the Texas Instruments CC2520 in the .INI file comments as well.

So I thought it would be best to explicitly showcase which parameters have been set in the simulation .INI file so whoever is using it can customize as needed.

src/inet/routing/leach/Leach.ned
```diff
@@ -0,0 +1,38 @@
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+//
+
+package inet.routing.leach;
+
+import inet.routing.contract.IManetRouting;
+
+//
```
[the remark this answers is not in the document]
n-jay: Resolved in https://github.com/inet-framework/inet/pull/929/commits/bb14ffbec782703cf3dd5506cf978c447aec111c

rhornig:
@CodiumAI-Agent /ask Is everything properly documented?

src/inet/routing/leach/Leach.cc
```diff
@@ -0,0 +1,450 @@
...
+        host = getContainingNode(this);
+        interfaceTable.reference(this, "interfaceTableModule", true);
+
+        clusterHeadPercentage = par("clusterHeadPercentage");
+        numNodes = par("numNodes");
+
+        roundDuration = dblrand(0) * 10;
+        TDMADelayCounter = 1;
+        event = new cMessage("event");
+        roundDuration = dblrand(0) * 10;
```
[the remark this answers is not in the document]
n-jay: Resolved in https://github.com/inet-framework/inet/pull/929/commits/c73ecaec2a66087f07d593546eb8221c10df40a4

rhornig:
Finally we did our homework and looked into the code-base (apart from my previous formal comments) and also went through the referenced paper. 

My most important question: What is the reason you want to integrate this PR into INET? There are valid reasons to add new code to INET, but we must be aware of the goal before we go further. 

At the moment, the code is not generic enough to be used by an INET user and lacks several essential features described in the paper. 

- To make sure, that it works correctly, it must be validated either against the paper's results (by reproducing their results) or against an other open source implementation (carrying out the same experiments).
- The application data must be generated by a user installable application. Right now, it inserts some ad-hoc temperature and humidity measurements inside the routing module. The data content, size and frequency must be user configurable. (i.e. it must work with any UDP application already available in INET)
- Important and meaningful statistics must be gathered by the routing module (for example the location of dying nodes, aggregation, cluster membership) 
- Data aggregation must be implemented (preferably as a separate, user configurable module). Without aggregation at cluster heads, each packet sent by a node will reach the base station sooner or later (in two hops: node->CH->BS) which means that the total energy consumption of the network will be definitely higher than in the case when each node just sends the packet directly to the base station (node->bs). This renders the whole protocol meaningless.
- There is no code that lowers the transmission power when sending to cluster heads, compared to the power used to send data from CH to BS. It means that nodes will not spare energy when transmitting to cluster heads as all transmission is using a power that reaches the BS anyway.
- The code seemingly does not switch the MAC to CDMA and does not use separate code words to avoid the collision between adjacent clusters. Not using code division multiple access would generate interference between neighboring clusters even if the transmission power is lowered.

n-jay:
Hi @rhornig, I appreciate this feedback very much. The answers to your questions are as follows:

> What is the reason you want to integrate this PR into INET?

LEACH is a popular and seminal routing protocol that is still being cited despite the original protocol being proposed over 20 years ago as seen from its many [variants](https://www.sciencedirect.com/science/article/abs/pii/S1570870520307356). 
While researching at the start of the project, I found a distinct lack of open-source LEACH simulation models that were up to date with recent versions of their respective simulation tools, if the code was made available at all that is. An example is [Solar-LEACH](https://omnetpp.org/download-items/SolarLEACH.html) from the OMNET model catalog that was submitted in 2006 for OMNET v3. This has also been mentioned by others in the community as per the [responses](https://groups.google.com/g/omnetpp/c/C7rSngRsT1M/m/c2-TfQmUBgAJ) I've received in the Google Group. 
Several people who were having the same problem and were searching for this code also reached out to me on my personal [repo](https://github.com/Agr-IoT/LEACH/issues?q=is%3Aissue+is%3Aclosed).

Since INET already has several multi-hop routing protocols and the community requirement I've observed about a LEACH simulation model, I thought it would be beneficial to contribute LEACH directly to the INET framework which can be further refined with feedback from the project Maintainers rather than just existing as a subpar community model.

> To make sure, that it works correctly, it must be validated either against the paper's results (by reproducing their results) or against an other open-source implementation (carrying out the same experiments).

As mentioned above, I've struggled against this myself as I couldn't find a suitable existing simulation hence why I wrote this OMNET model from scratch. There seems to be a community-developed MATLAB [project](https://in.mathworks.com/matlabcentral/fileexchange/66574-leach), however that seems be a partial implementation.

> At the moment, the code is not generic enough to be used by an INET user and lacks several essential features described in the paper.

Noted. 
Is it possible to develop these features in the same PR? I would very much like to continue working on this and add the remainder with your feedback when it comes to aspects like the data aggregation module and statistic collection.

rhornig:
Great. If your goal is to create a model that could be used by other community members and you are willing to put the effort into this, we could proceed with this PR. I can assist you to achieve this (though obviously I cannot guarantee the end result). We kind of went the wrong way starting this PR by discussing formal stuff and implementation details first, but first general concepts should be discussed, so here is my recommendation:

1. discuss the main goal and scope of the project
2. define what features are needed / missing / important
3. design and implementation strategy for the features
4. implementation details, formalities

So I start with 1. Here is my recommendation: Pick a version of the article (you already did this) and set a goal to reproduce the results of that article (this is in itself a worthy and publishable goal). So in this case, we want to reproduce all charts and statistics the original paper provided. This can be seen as a way to validate the model and also could be used as a baseline for other people who want to develop new protocols. It would also give guidance about necessary statistics, configuration parameters etc. Let's say we want to compare direct NODE to BaseStation communication's to Leach. Compare the energy usage and the battery depletion distribution over time / space. The current paper already defines that.

What do you think?

n-jay:
Sounds good @rhornig! 
I consider this a research project, so I'm perfectly fine with this result-based approach with your mentorship. 😄 

> So in this case, we want to reproduce all charts and statistics the original paper provided. This can be seen as a way to validate the model and also could be used as a baseline for other people who want to develop new protocols.

Understood. I assume this implementation is to be done with the inclusion of the generic model aspects you mentioned in the earlier comment (user installable application etc), yes?
I will quickly resolve the formal implementation fixes you suggested and then move into adding the generic components and statistic gathering.

> but first general concepts should be discussed, so here is my recommendation:

Noted. 
Additionally, do you think it would be best to maintain separate correspondence via a dedicated mail thread perhaps? 
I felt so as this project has sort of expanded beyond just this PR, so adding all replies here could end up spamming this PR discussion, which can be reserved solely when code changes are involved.

src/inet/routing/leach/Leach.ned
```diff
@@ -0,0 +1,38 @@
...
+        // properties
+        @class("leach::Leach");
+        @display("i=block/routing");
+        
+        // context parameters
+        string interfaceTableModule; 
+        
+        // LEACH parameter group
+        double maxVariance = default(1);
+        int numNodes = default(10);
```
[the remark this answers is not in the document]
n-jay: Resolved in https://github.com/inet-framework/inet/pull/929/commits/e99fabbf6bce1697931192c565820324118d3ca0

src/inet/routing/leach/Leach.ned
n-jay: > Generally, if you specify a default value in a NED file, then that value must be supported by the specification document somewhere.

Noted. 
The original paper authors have provided 0.05 (or 5%) as the **desired percentage of cluster heads** for their experiments so took your recommendation and modified the code in https://github.com/inet-framework/inet/pull/929/commits/f04b5be512cb34a55b92b7eacf6856c6e85ddc0d

src/inet/routing/leach/Leach.cc
n-jay: Understood. Resolved in https://github.com/inet-framework/inet/pull/929/commits/ad80f49543c129b25da92bfee6b18b9a49c2a0d7

src/inet/routing/leach/Leach.cc
n-jay: Noted, sorry about that. Seems to be a remnant from when I was testing with multiple self message types.
Resolved in https://github.com/inet-framework/inet/pull/929/commits/2a7da9283b9ece61cc0a94d03fb3902cc4ec35d0

src/inet/routing/leach/LeachBS.cc
n-jay: Resolved in https://github.com/inet-framework/inet/pull/929/commits/8dcc7d590e6952bca4f714c7611c13a4928a3161

src/inet/routing/leach/Leach.h
n-jay: I looked into this and it does seem that vectors
So something like `unordered_map<Ipv4Address, nodeMemoryObject> nodeMemory;` yes?

With a modified nodeMemory struct looking something like this to avoid having redundant nodeAddress value.
```
struct nodeMemoryObject {
        Ipv4Address CHAddr;
        double energy;
};
```

rhornig: Yes, but this is really just optimization so it can be done at the end of the project. Use whichever structure is more natural to you.

src/inet/node/leach/LeachNode.ned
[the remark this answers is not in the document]
rhornig: okay. I get this.

src/inet/routing/leach/Leach.h
n-jay: Understood. I'll keep it as it is for now because it seems like I'll have to change several other components to accommodate this. So will keep this suggestion open to take care of in future.

I'm now moving on to integrating the UDP app with the LEACH module for data packet transmission.
Will update below.

n-jay:
> i.e. it must work with any UDP application already available in INET)

@rhornig, I'm currently working on this and need a bit of clarification. Is it required that _all_ packet types (control, acknowledgment, etc.) used in LEACH be UDP packets or just the packets with sensed data?

I feel like the former (all packets UDP) is the way to go because otherwise, it's going to get confusing when filtering packets in the `processMessage` method. 
I can drop the `setPacketType` use and instead use the `setKind` to perform the filtering. 
Let me know if this is the ideal approach.

n-jay:
@rhornig, I implemented UDP packets in https://github.com/inet-framework/inet/pull/929/commits/6008a9a3634e3c9dc9ba18ac357da8dc9adec5ec.
Appreciate some feedback on whether this approach is correct so I can replicate it for all packet transmissions
https://github.com/inet-framework/inet/blob/6008a9a3634e3c9dc9ba18ac357da8dc9adec5ec/src/inet/routing/leach/Leach.cc#L282
https://github.com/inet-framework/inet/blob/6008a9a3634e3c9dc9ba18ac357da8dc9adec5ec/src/inet/routing/leach/Leach.cc#L358

n-jay:
Hi @rhornig,
I started finalizing the UDP packet support, but an issue with my test simulation parameters is currently holding me back.
Instead of just writing it here, I've created a new question in the discussion section to benefit others: https://github.com/inet-framework/inet/discussions/976.

I would really appreciate some feedback on this so I can quickly finalize the rest of the protocol.

n-jay:
Hi @rhornig

> I started finalizing the UDP packet support, but an issue with my test simulation parameters is currently holding me back.
Instead of just writing it here, I've created a new question in the discussion section to benefit others: https://github.com/inet-framework/inet/discussions/976.

I made multiple attempts to solve the simulation issue based on the showcases, tutorials and examples in INET but had no luck.
I even checked out to my code to one of earliest commits (https://github.com/inet-framework/inet/pull/929/commits/67c97be16b8eaf83466d63ab1ff8e4fb54cb9cfa) assuming the issue must be resulting from a change I've made based on the comments, but made no difference. The same codebase works fine in INET 4.2.5 (on OMNET v5).

I set debug points and identified that the CH election in the protocol works as expected, reaching the packet transmission point without any exceptions or issues:
https://github.com/n-jay/inet/blob/f4a4040557663d453a91fb43b81282c27cef8166/src/inet/routing/leach/Leach.cc#L307

But the issue is that the Non-CH nodes seem to be receiving the sent packets, and the debug points set within the `processMessage` method are not triggered.
https://github.com/n-jay/inet/blob/f4a4040557663d453a91fb43b81282c27cef8166/src/inet/routing/leach/Leach.cc#L130

Would greatly appreciate some feedback on this because this is a major blocker preventing me from finalizing the protocol.

n-jay:
> Would greatly appreciate some feedback on this because this is a major blocker preventing me from finalizing the protocol.

Happy to inform that the issue is resolved, @rhornig. Described the fix in the discussion comment: https://github.com/inet-framework/inet/discussions/976#discussioncomment-9955669.
Will now proceed to finalize the rest of the protocol quickly and update you here.

PR #973, opened 2024-05-07
asynchronous traffic shaper with scheduler groups

terlue:
Current implementation of the ATS algorithm in EligibilityTimeMeter sets groupEligibilityTime per stream, this can break the FIFO order of packets in the switch (See: IEEE 802.1Qcr, Section 8.6.5.6 (NOTE 2), and Specht, J., and S. Samii, “Urgency-Based Scheduler for Time-Sensitive Switched Ethernet Networks,” 28th Euromicro Conference on Real-Time Systems (ECRTS), pp. 75–85, 2016. ) . 

This change introduces a ATSIeee8021qFilter with a setup of a GroupEligibilityTimeTable, GroupEligibilityTimeMeter, and EligibilityTimeFilter, to replace the current ATS setup using the EligibilityTimeMeter and EligibilityTimeFilter modules. 
Packets arriving at the GroupEligibilityTimeMeter are checked for their group (ingress port + priority), if there already is a groupEligibilityTime in the table, this is used as groupEligibilityTime in the algorithm, otherwise the new group and the eligibility time of this packet are added to the table.

### DIFF
--- a/src/inet/linklayer/ieee8021q/ATSIeee8021qFilter.ned
+++ b/src/inet/linklayer/ieee8021q/ATSIeee8021qFilter.ned
@@ -1,0 +1,32 @@
+//
+// Copyright (C) 2020 OpenSim Ltd.
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+//
+
+package inet.linklayer.ieee8021q;
+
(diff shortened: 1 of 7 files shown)


levy:
Thank you for submitting this PR! It looks good to me, so I merge this into master.

PR #983, opened 2024-06-26
NeighborListCache: don't build neighbor list during initialization

chenlijun99:
Given n radios, neighbor list construction has O(n^2) complexity. If, during initialization, when each radio is added, the neighbor list is reconstructed, we end up having a simulation initialization with O(n^3) complexity.
Sidenote: In one of our tests, without this patch, when running a WiFi simulation of 5000 nodes, it took ~40 minutes to just initialize the simulation.


levy:
Thanks for submitting this PR!

PR #1013, opened 2024-12-12
Fix ChunkSerializerRegistry type_info matching on Windows

Musteblume:
Fixes error that the registry is not finding serializers on Windows.

type_info addresses and names are not guaranteed to be unique. To my knowledge before() is the only reliable way to compare type_info. This worked on Windows 11 24H2 and Ubuntu 24.04 for me.


ZoltanBojthe:
Levente Meszaros added a fix for my commit:
2116a6ad1956
ChunkSerializerRegistry: Fixed missing header file caused breaking compilation on Windows and MacOS.

Issue fixed.

PR #1104, opened 2026-05-31
[FIX] [IPv6] Use correct destination address so packets with multicast destination are properly handled

mgonzalezlopezudc:
The previous version was inefficient because it used the broadcast address as the destination address for any multicast packet.

### DIFF
--- a/src/inet/networklayer/ipv6/Ipv6.cc
+++ b/src/inet/networklayer/ipv6/Ipv6.cc
@@ -403,7 +403,10 @@ void Ipv6::datagramLocalOut(Packet *packet, const NetworkInterface *destIE, Ipv6
     const auto& ipv6Header = packet->peekAtFront<Ipv6Header>();
     // route packet
     if (destIE != nullptr)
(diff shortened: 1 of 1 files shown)


avarga:
Part of `topic/ipv6-refactoring`, will be merged into `master` soon.

avarga:
Thanks for this fix, @mgonzalezlopezudc — the diagnosis is spot on. IPv6 multicast datagrams were being handed to the link layer with the broadcast MAC instead of the proper `33:33:xx:xx:xx:xx` multicast MAC mapped from the destination address.  Based on your input, the following fixes landed on `master`, as part of the recent IPv6 multicast forwarding work:

- `datagramLocalOut()` — the multicast-MAC check is now in place (commit 3be7618b37, *"IPv6: fix: use correct multicast MAC instead of broadcast (RFC 2464 Section 7)"*), matching your change with the `else` branch wrapped in braces.
- `routeMulticastPacket()` — superseded by a larger refactor that also uses `mapToMulticastMacAddress()` (commit dd56c312b2, *"IPv6: implement multicast forwarding (RIB/FIB + RPF)"*).

PR #1109, opened 2026-06-22
Add lifecycle support for BGPv4 routers

giovanninardini:
This PR adds lifecycle support to the BGPv4 model so BGP routers can participate in INET node shutdown, crash, and startup operations.
The implementation converts `Bgp` to the lifecycle-aware routing protocol pattern, removes local BGP-learned IPv4 routes when the module stops, closes or aborts active BGP/TCP sessions depending on whether the operation is graceful or crash-like, and recreates BGP router/session state on startup. Restarted lifecycle-capable nodes reenter the existing RFC 4271 FSM through normal start events, allowing sessions to be reestablished and routes to be learned again after a node resumes.
This also adds a new `examples/bgpv4/BgpLifecycle` scenario that exercises shutdown, restart, and crash behavior over a longer run, plus persistent fingerprint coverage for that scenario.

**Changes**

- Make `Bgp` lifecycle-aware instead of lifecycle-unsupported.
- Share normal initialization/start behavior through `startBgp()`.
- Add stop/crash cleanup paths:
  - cancel pending startup;
  - remove local IPv4 routes with source type`IRoute::BGP`;
  - close sessions on graceful stop;
  - abort sessions on crash;
  - delete the current `BgpRouter`.
- Recreate fresh BGP router/session/timer/socket state on startup.
- Handle peer TCP close/failure so lifecycle-capable BGP nodes can reattempt normal FSM startup.
- Add `BgpLifecycle` example under `examples/bgpv4`
- Add `BgpLifecycle` to `tests/fingerprint/examples.csv`

**Validation**

- Existing BGP smoke test: examples/bgpv4/BgpUpdate
- Lifecycle smoke tests:
  - shutdown removes local BGP route;
  - startup reestablishes BGP sessions and exchanges UPDATEs again;
  - crash scenario completes without error.
- All BGP-filtered fingerprints PASS

**Notes**

Crash remains non-graceful: the implementation does not send explicit BGP NOTIFICATION or withdrawals during crash. Graceful shutdown closes the TCP sessions and removes local BGP routes.

---

<a href="https://app.devin.ai/review/inet-framework/inet/pull/1109" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


avarga:
Thanks for this contribution, @giovanninardini - really nice, well-scoped work, and the `BgpLifecycle` smoke test made it easy to validate. Merged into `master`.

For the record, here is how it landed and what was verified:

**Integration**
- The five commits were reworded to INET's `BGP: …` commit-message convention (authorship and dates preserved), and applied on top of current `master`.

**Verification**
- Release build OK.
- All bgpv4 fingerprint tests pass: the 7 pre-existing rows are byte-identical (no regression), and the new `BgpLifecycle` row reproduces your recorded fingerprint exactly on a fresh build.
- Behavior confirmed end-to-end: shutdown removes the locally-learned BGP route; restart re-establishes the session and re-learns routes via UPDATE; crash tears down cleanly and the run completes.

**Follow-up commit added on top**
- `isLifecycleNode()` now resolves the node via `getContainingNode()` instead of `getParentModule()`, and carries a comment documenting *why* the FSM only re-arms reconnection on lifecycle-capable nodes. That gating turns out to be load-bearing: enabling auto-reconnect unconditionally re-opens a listening socket on an already-bound port and crashes the plain (non-lifecycle) BGP scenarios — so it should stay gated.

**Deferred (separate follow-up)**
- Making graceful stop *fully* graceful — i.e. keeping the node up briefly so TCP can flush the teardown before the sessions are destroyed (the pattern RIP uses via `delayActiveOperationFinish()`) — surfaces a latent reconnection issue: while a session is still closing, a lifecycle-capable peer reconnects and re-binds port 179 (`Address already in use`). That's the same socket-cleanup-on-reconnect gap as above, and fixing it properly is its own change. The current synchronous close (no NOTIFICATION/withdrawals) is therefore retained, as you intended.

PR #1110, opened 2026-06-30
Bgp: add Adj-RIB-In support and route withdrawal failover

giovanninardini:
## Summary

This PR adds Adj-RIB-In based route tracking to the BGP implementation so routes learned from peers are retained separately from the selected Loc-RIB route. When a peer withdraws a prefix or a BGP session fails, the router now removes only the affected learned paths, reruns best-path selection for those prefixes, and either installs an alternate route or advertises a withdrawal.

## Details

- Add an Adj-RIB-In table to `BgpRouter` for all peer-learned routes.
- Rework UPDATE processing to:
  - handle multiple IPv4 NLRI entries,
  - handle IPv6 MP_REACH_NLRI and MP_UNREACH_NLRI,
  - track affected prefixes,
  - rerun best-path selection only for changed prefixes.
- Add route source metadata to BGP route entries, including source session and peer address.
- Add explicit BGP withdrawal sending for IPv4 and IPv6.
- Withdraw routes learned from a session when established sessions fail or timers expire.
- Guard UPDATE, KEEPALIVE, and WITHDRAW sends when the TCP socket is no longer connected.
- Stop BGP session timers when the FSM transitions back to Idle.
- Restart BGP after lifecycle startup through the timer path so interfaces can come back up first.
- Add a lifecycle stage for network interface configuration and move IPv6 interface reconfiguration before IPv6 address assignment.
- Bind outgoing BGP TCP sockets to the configured local BGP address.
- Ignore stale TCP messages for sockets that have already been replaced after restart/failure.

## Examples

Adds two new diamond-topology examples that demonstrate failover, failback, and a second failure:

- `examples/bgpv4/BgpDiamondFailover`
- `examples/bgpv4/BgpDiamondFailover6`

Both examples prefer the `A-B-C` path initially, fail over to `A-D-C` when router `B` goes down, fail back after `B` restarts, and then exercise failover again after a crash.

## Testing

- Updated existing BGP fingerprints affected by the new route-selection behavior.
- Added fingerprint coverage for the new IPv4 and IPv6 diamond failover examples.

---

<a href="https://app.devin.ai/review/inet-framework/inet/pull/1110" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/examples/bgpv4/BgpDiamondFailover/BGPConfig.xml
+++ b/examples/bgpv4/BgpDiamondFailover/BGPConfig.xml
@@ -1,0 +1,44 @@
+<?xml version="1.0" encoding="ISO-8859-1"?>
+<BGPConfig xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="BGP.xsd">
+
+    <AS id="65001">
+        <Router interAddr="10.0.1.1">
+            <Network address="10.0.1.0"/>
+        </Router>
+    </AS>
+
+    <AS id="65002">
+        <Router interAddr="10.0.12.2"/>
+    </AS>
+
+    <AS id="65003">
+        <Router interAddr="10.0.3.1">
+            <Network address="10.0.3.0"/>
+        </Router>
+    </AS>
+
+    <AS id="65004">
+        <Router interAddr="10.0.14.4"/>
+    </AS>
+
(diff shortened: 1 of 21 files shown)


avarga:
Merged manually. Thanks for your help in making INET's BGP a more complete and robust model!

PR #1128, opened 2026-08-15
Add HT compressed Block Ack support

mgonzalezlopezudc:
## Summary

Add one-TID HT Compressed Block Ack support to the IEEE 802.11 QoS MAC, including originator selection, recipient bitmap construction, wire encoding, and the complete BAR/BA exchange. Basic Block Ack remains the default.

> [!IMPORTANT]
> **Depends on #1146 (HT Greenfield) and #1148 (ADDBA transaction and agreement lifecycle).** This branch is now rebased onto `cleanup/feat-ht-greenfield` at `6935375744`, with the ADDBA prerequisite commits retained above it. Merge both prerequisites before merging this PR, then rebase onto `master`. Greenfield and ADDBA prerequisite changes belong to their respective PRs.

## Changes

- Correct Basic and Compressed BAR headers to 20 bytes before FCS, represent Compressed BA with a 28-byte header, and encode BAR/BA Control and Starting Sequence Control fields in little-endian wire format.
- Route Basic and Compressed variants through HCF, frame sequences, acknowledgment processing, rate selection, protection timing, and receive reordering. Predict response rates using the correct peer for outgoing and received BARs while preserving #1146's response-rate policy.
- Generate the one-TID 64-bit compressed bitmap, including an all-zero response when matching recipient partial state is absent. Restrict compressed selection to eligible immediate agreements and unfragmented MPDUs.
- Keep the recipient acknowledgment window authoritative, preserve leading holes and cyclic delivery order, and make receive-buffer admission transactional. Record frames at the Block Ack window start and centralize cyclic receive-window arithmetic.
- Preserve Block Ack transaction state and correlate responses with the active receive step.
- Reject ineligible BARs before changing receive state or renewing agreement lifetime. Preserve #1148's active-agreement checks, generation and inactivity handling, and receive-lifetime behavior.

## Capability and architecture

Compressed selection is explicitly configured on `OriginatorBlockAckAgreementPolicy` through `localCompressedBlockAckSupported` (default `false`) and `compressedBlockAckPeerAddresses` (default empty). Both local support and the configured peer address are required. This models explicit per-peer support; it does not add automatic negotiation of compressed Block Ack capability.

The feature touches IEEE 802.11 frame definitions and serialization, Block Ack agreement and acknowledgment-policy contracts, HCF dispatch, reordering, and rate/protection calculations. No feature descriptors or sealed source paths are changed, and no new architecture or naming exceptions are introduced.

[...]

PR #1136, opened 2026-08-20
Ieee80211: add support for HT primary and secondary channels (HT40- and HT40+)

mgonzalezlopezudc:
> [!IMPORTANT]
> Depends on #1128 (HT compressed Block Ack support). Review and merge #1128 before this PR.

### Summary of Changes

This PR implements support for IEEE 802.11n High Throughput (HT) 40 MHz primary and secondary channel operation (HT40- and HT40+), multi-threshold subchannel Clear Channel Assessment (CCA), and EDCA secondary channel DIFS idle sensing and backoff restart per IEEE Std 802.11-2024.

#### 1. Physical Layer & Channel Geometry
- **Ieee80211Channel**: Added `Ieee80211SecondaryChannelOffset` enum (`none`, `above`, `below`) and center frequency calculation methods (`getSecondaryCenterFrequency()`, `getBondedCenterFrequency()`, `getSecondaryChannelNumber()`) per Table 9-134.
- **Ieee80211Transmitter**: Transmits on bonded center frequency for 40 MHz HT modes and on primary center frequency for 20 MHz legacy/HT20 frames.
- **Ieee80211Receiver**: Added multi-threshold HT CCA sensing on 20 MHz listening slices (-82 dBm HT20/OFDM, -79 dBm HT40 bonded, -62 dBm energy detection) per Clause 19.3.19.6.
- **IIeee80211CcaProvider & Ieee80211Radio**: Implemented CCA provider interface and registered `ccaStateChangedSignal` with subchannel busy snapshots.
- **Background Noise**: Scaled power for 20 MHz listening slices during 40 MHz radio operation.

#### 2. MAC & MIB Layers
- **IRx / Rx**: Subscribed to CCA state changes, tracked primary and secondary channel idle durations (`isSecondaryChannelIdleFor()`), and reflected primary CCA status in medium free recomputation.
- **Hcf / Edcaf / Dcaf**: Implemented EDCA channel access restart on secondary DIFS busy (`SIFS + 2 * slotTime`) per Clause 11.15.9, retaining CW without bumping retry counters or dropping frames.
- **Ieee80211Interface & Ieee80211Mib**: Forwarded `htSecondaryChannelOffset`, `htShortGi40`, and `htMaxMcs` parameters.

#### 3. Verification & Tests
- **Unit Tests**:
  - `tests/unit/Ieee80211Ht40SecondaryChannel_1.test`: Channel geometry, offsets, and out-of-bounds error handling.
  - `tests/unit/Ieee80211HtCcaSensitivity_1.test`: Multi-threshold CCA sensitivities and mode set verification.
- **Example Simulation**:
  - `examples/wireless/channelwidths/ChannelWidthsNetwork.ned` & `omnetpp.ini`: Configs for `Ht20MHz`, `Ht40MHzSecondaryAbove`, `Ht40MHzSecondaryBelow`, and `Ht40MHzSecondaryAboveWithInterferer`.

---

<a href="https://app.devin.ai/review/inet-framework/inet/pull/1136" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


mgonzalezlopezudc:
@levy 

# Response to reviewer (Devin)

**Devin agrees**

The reviewer comments provide valuable validation and inquiries into the HT40 MAC/PHY architecture. The technical assessments demonstrate that:
1. **Critical defects and behavioral bugs identified during review cycles have been fixed and verified** (contention callback reentrancy, channel-band inheritance in `handleUpperCommand`, medium sensing during reception, band-aware DIFS/PIFS secondary idle checks, and sub-band SNIR noise scaling).
2. **Flagged items regarding standards behavior (e.g., HT40 EDCA backoff restart vs. 20 MHz fallback, HT multi-threshold CCA) are confirmed to be fully compliant with IEEE Std 802.11-2024**. Modifying the model to force dynamic fallback is **inconvenient and unnecessary**, as baseline Option (b) is normatively sound.
3. **Physical layer relaxations (background noise sub-band scaling, band containment in analog models, SNIR digital filter scaling) are physically rigorous**, eliminating artificial simulation crashes and modeling RF receiver passbands accurately.
4. **All unit tests, fingerprint regressions, and wireless throughput examples pass in debug mode (`MODE=debug`)**.

---

## Detailed Technical Responses to Reviewer Comments

### Section 1: FLAGS (In-Depth Technical Assessments)

---

#### 1. HT40 EDCA backoff restart can starve transmission when secondary stays busy
* **Location**: [Hcf.cc](src/inet/linklayer/ieee80211/mac/coordinationfunction/Hcf.cc#L230-L234) (`src/inet/linklayer/ieee80211/mac/coordinationfunction/Hcf.cc`), [Rx.cc](src/inet/linklayer/ieee80211/mac/Rx.cc#L171-L189) (`src/inet/linklayer/ieee80211/mac/Rx.cc`), [Ieee80211Radio.cc](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc#L66) (`src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc`)
* **Reviewer Comment**: `Hcf::channelGranted` calls `shouldRestartHt40ChannelAccess` and, when the secondary channel was not idle for the required DIFS/PIFS window, invokes `edcaf->restartChannelAccess` and returns without transmitting (`Hcf.cc:230-234`). The restart draws a fresh backoff with the same CW but never falls back to a primary-only 20 MHz transmission, so a persistently busy secondary channel indefinitely defers the frame. IEEE 802.11 permits transmitting a 20 MHz PPDU on the primary in this case; here throughput can stall entirely while the secondary is occupied (e.g. the `Ht40MHzSecondaryAboveWithInterferer` example). Also note that after every transmission the radio leaves HT40 receiver mode, resetting `secondaryCcaIdleSince`, which can force an extra restart on the next access.
* **Convenience & Disposition**: **VALID CRITIQUE & NUANCED STANDARDS EVALUATION — DOCUMENTED IN FULL FIDELITY; CONFIRMED STANDARDS-COMPLIANT OPTION (B) BASELINE WITH NOTED SIMULATION RESTART CHARACTERISTICS**.
* **Detailed Technical Assessment**:

  ##### A. Starvation Under Persistent Secondary Interference (Option b vs. Option a)
  - **IEEE Std 802.11-2024 Normative Specification**: Clause 11.15.9 (*"40 MHz channel access"*, `80211ax-2024:chunk:06337`, p. 2639) states:
    > *"If a STA was unable to transmit a 40 MHz mask PPDU because the secondary channel was occupied during this interval, it may take one of the following steps:*
    > *a) Transmit a 20 MHz mask PPDU on the primary channel.*
    > *b) Restart the channel access attempt. In this case, the STA shall invoke the backoff procedure as specified in 10.23.2 as though the medium is busy as indicated by either physical or virtual CS and the backoff counter has a value of 0."*
  - **Standard Compliance of Option (b)**: The standard explicitly provides Option (b) as a valid normative path. Under Option (b), backoff is re-invoked using the current $CW[AC]$ without incrementing retry counters. When the secondary channel is *continuously* occupied by interference, Option (b) repeatedly restarts the backoff procedure, deferring the 40 MHz transmission indefinitely. This throughput stall (starvation of 40 MHz frames) is the exact expected physical and normative consequence of choosing Option (b) in a permanently congested secondary channel environment.
  - **Option (a) as a Future Extension**: Option (a) requires dynamic, on-the-fly PHY/MAC fallback (recomputing PPDU framing, MCS, preamble format, MPDU/A-MPDU length, and NAV/TXOP duration at the exact instant of channel grant). This represents dynamic bandwidth rate adaptation, which is a feature extension beyond baseline HT40 channel access.

  ##### B. Secondary Idle Timer Reset and Clock Decoupling Analysis
  - **The Mechanism**: In [Ieee80211Radio.cc:66](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc#L66), `ht40` CCA is gated on `isReceiverMode(radioMode)`. While transmitting, `ht40` is reported as `false`. Upon returning to receiver mode, `ccaStateChangedSignal` fires with `ht40 = true`. In [Rx.cc:185-188](src/inet/linklayer/ieee80211/mac/Rx.cc#L185-L188), `!wasHt40Cca` evaluates to `true`, unconditionally resetting `secondaryCcaIdleSince = simTime()` to the instant of receiver entry ($T_{\mathrm{rx\_enter}}$).
  - **Independent Clocks & Spurious Extra Restart**:
    1. `Contention` operates on discrete slot increments driven by primary channel state (`mediumFree`), while `Rx::isSecondaryChannelIdleFor` measures continuous wall-clock time (`simTime() - secondaryCcaIdleSince`).
    2. Because `secondaryCcaIdleSince` is reset to $T_{\mathrm{rx\_enter}}$, any secondary channel idle history that accumulated *prior* to the transmission is discarded upon RX re-entry.
    3. If a station obtains channel access at time $t < T_{\mathrm{rx\_enter}} + \mathrm{requiredIdle}$ (for example, with a very small backoff draw, or if contention started before the transition settled), `isSecondaryChannelIdleFor(requiredIdle)` will evaluate to `false`, triggering an extra backoff restart even if the physical secondary channel was completely clear before, during, and after the transmission.
    4. On the subsequent backoff restart, the station draws a fresh backoff counter; during this new contention phase, elapsed time will easily exceed $\mathrm{requiredIdle}$ (provided the secondary channel remains clear), allowing the transmission to proceed.
  - **Architectural Assessment & Modeling Alternatives**:
    - **Current Implementation as a Known Modeling Simplification**: The current reset on `!wasHt40Cca` represents a conservative modeling simplification where leaving receiver mode invalidates the secondary CCA idle tracker. This is fully compliant with Option (b) backoff rules, but the resulting extra restart is an implementation timing artifact rather than a standards-mandated behavior.
    - **Preserving Pre-TX Idle History (Non-Unphysical Alternative)**: As noted in review, eliminating this spurious restart does *not* require unphysical in-flight channel sensing during transmission. A clean, physically sound modeling alternative is to avoid clobbering an already valid pre-TX idle timestamp upon RX re-entry. In this pattern, the initial `if (!ht40Cca || secondaryCcaBusy) secondaryCcaIdleSince = -1;` check is preserved first (ensuring history is immediately invalidated whenever the secondary is genuinely busy or HT40 CCA is inactive), while the `else if` branch only stamps `secondaryCcaIdleSince = simTime()` if `secondaryCcaIdleSince < 0` or `wasSecondaryBusy`. This preserves pre-TX idle history across transmissions without sensing during TX.
    - **Scope Verdict**: Leaving the current reset behavior is a defensible scope decision for baseline HT40 channel access, with the understanding that preserving pre-TX idle timestamps provides a well-defined enhancement path if higher throughput measurement fidelity under secondary-channel contention is required.

---

#### 2. CCA snapshot is refreshed only on transceiver-state transitions
* **Location**: [Ieee80211Radio.cc](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc#L58-L86) (`src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc`), [RadioMedium.cc](src/inet/physicallayer/wireless/common/medium/RadioMedium.cc#L216-L222) (`src/inet/physicallayer/wireless/common/medium/RadioMedium.cc`), [MediumLimitCache.cc](src/inet/physicallayer/wireless/common/medium/MediumLimitCache.cc#L152-L188) (`src/inet/physicallayer/wireless/common/medium/MediumLimitCache.cc`)
* **Reviewer Comment**: `Ieee80211Radio::updateTransceiverState` calls `updateCcaState` (`src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc:82-86`), and `updateTransceiverState` is invoked by the base radio on reception/transmission start/end and radio-mode changes. This means per-channel busy/idle for the secondary channel is only re-evaluated at those transition points, computing an instantaneous `listenOnMedium` over a 1-raw-tick window (`computeIsBandBusy`, `Ieee80211Radio.cc:46-56`). This should coincide with the moments interference actually starts/stops (each neighbor reception triggers a reception-state update), so the secondary CCA idle/busy tracking in `Rx` stays in sync. Flagging so a reviewer confirms no interference change can occur without a corresponding transceiver-state update that would leave the snapshot stale.
* **Convenience & Disposition**: **CONFIRMED OPTIMAL AS-IS FOR DEFAULT/SHIPPED CONFIGURATIONS; RANGE-FILTER BOUNDARY CONDITION PRECISELY BOUNDED & DOCUMENTED**.
* **Technical Rationale & Range Filtering Analysis**:
  - **Discrete-Event Physical Layer Semantics**: In INET's discrete-event simulation model, physical medium state and interference levels are piecewise-constant. Channel power changes *strictly and exclusively* when a radio starts transmitting, finishes transmitting, or changes operating mode.
  - **Synchronous Event Propagation**: Whenever any transmitter in the medium begins or ends a transmission, `RadioMedium` immediately notifies all receiving radios via `receptionStarting` and `receptionEnding`, which in turn invoke `Radio::updateTransceiverState()`.
  - **Default Range Filter (`rangeFilter = ""`)**: By default, `RadioMedium.rangeFilter` is disabled (`""`), meaning `RadioMedium::isInInterferenceRange` returns `true` for all transmissions. Every transmission arrival/departure triggers `updateTransceiverState()` on all radios, guaranteeing zero staleness for `updateCcaState()` with zero timer polling overhead.
  - **Non-Default `rangeFilter` Boundary Condition**:
    1. In `MediumLimitCache::computeMinInterferencePower`, `Ieee80211Receiver` inherits `ReceiverBase::getMinInterferencePower()`, which returns `NaN` (consistent with legacy 802.11 receiver behavior).
    2. Therefore, maximum interference range is governed entirely by the medium-level parameters (`minInterferencePower` / `maxInterferenceRange`).
    3. If a user explicitly enables `rangeFilter = "interferenceRange"` or `"communicationRange"` AND configures a finite medium cutoff less sensitive than $-82\mathrm{ dBm}$ (the HT CCA sensitivity), a distant transmitter exceeding that geometric cutoff but arriving above $-82\mathrm{ dBm}$ at the receiver would be filtered before reaching `receptionStarting`, potentially leaving the CCA snapshot stale.
  - **Caveat & Conclusion**: In all standard, example, and default configurations (`rangeFilter = ""`), the CCA snapshot is strictly non-stale and optimal without polling. For advanced non-default setups using spatial range filters, the medium's `minInterferencePower` must be set at least as sensitive as the HT CCA threshold ($-82\mathrm{ dBm}$) to avoid premature geometric cutoff of secondary interferers.

---

#### 3. Channel reconfiguration constructs a validating `Ieee80211Channel` that can throw
* **Location**: [Ieee80211Radio.cc](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc#L146-L152) (`src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc`), [Ieee80211ControlInfo.msg](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211ControlInfo.msg#L18-L26) (`src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211ControlInfo.msg`), [Ieee80211MgmtSta.cc](src/inet/linklayer/ieee80211/mgmt/Ieee80211MgmtSta.cc#L206-L215) (`src/inet/linklayer/ieee80211/mgmt/Ieee80211MgmtSta.cc`)
* **Reviewer Comment**: In `Ieee80211Radio::handleUpperCommand`, a bitrate/channel-only reconfigure inherits the radio's current `htSecondaryChannelOffset` (`targetSecondaryChannelOffset = ... : htSecondaryChannelOffset`) and validates by constructing `Ieee80211Channel(targetBand, targetChannelNumber, targetSecondaryChannelOffset)`. The `Ieee80211Channel` constructor eagerly resolves the secondary center frequency and throws `cRuntimeError` if the secondary channel falls outside the band. If a radio operating in HT40 (offset above/below) is asked to switch to a band-edge primary channel via a plain channel-number command, this would abort the simulation rather than fail gracefully.
* **Convenience & Disposition**: **VALID LATENT DEFECT / KNOWN SCOPE LIMITATION — FULLY CONCEDED AND PRECISELY DOCUMENTED (RESOLUTIONS IDENTIFIED FOR DYNAMIC HT40 AGENTS)**.
* **Detailed Technical Assessment & Root Cause**:
  - **The Gap Identified by Reviewer**:
    1. [Ieee80211ControlInfo.msg:18-26](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211ControlInfo.msg#L18-L26) defines `Ieee80211ConfigureRadioCommand`, which carries `channelNumber` and `channel`, but has no scalar `secondaryChannelOffset` field.
    2. Upper-layer scanning agents like [Ieee80211MgmtSta::changeChannel](src/inet/linklayer/ieee80211/mgmt/Ieee80211MgmtSta.cc#L206-L215) issue plain channel-number-only commands: `configureCommand->setChannelNumber(channelNum)`.
    3. When an HT40 radio (with `htSecondaryChannelOffset` set to `above` or `below`) receives a channel-number-only command, `Ieee80211Radio::handleUpperCommand` inherits the retained `htSecondaryChannelOffset` and constructs `Ieee80211Channel(targetBand, targetChannelNumber, targetSecondaryChannelOffset)`.
    4. If the destination channel is at a band edge (e.g. primary channel 10 with offset `above` in an 11-channel 2.4 GHz band), the `Ieee80211Channel` constructor throws `cRuntimeError("Secondary channel frequency ... falls outside band")`, aborting the simulation.
  - **Practical Severity & Scope**:
    - In practice, this latent defect is not triggered in the shipped HT40 examples because they operate on fixed channels and disable STA management agents (`staAgent.typename = ""`).
    - Dynamic channel switching / scanning under HT40 operation is currently out of scope for the baseline implementation.
  - **Clear Enhancement & Resolution Paths for Dynamic Agents**:
    - **Option (a) Graceful Radio Fallback (Recommended)**: In `Ieee80211Radio::handleUpperCommand`, when handling a channel-number-only command where the inherited `htSecondaryChannelOffset` would exceed the band boundaries, automatically fall back to `IEEE80211_SECONDARY_CHANNEL_NONE` (with an `EV_WARN` log) instead of allowing the constructor to throw.
    - **Option (b) Message API Expansion**: Add `secondaryChannelOffset` to `Ieee80211ConfigureRadioCommand.msg` and update `Ieee80211MgmtSta::changeChannel` so upper-layer agents can explicitly control channel width and secondary offset during scanning.

---

#### 4. HT CCA listening-decision override also changes plain HT20 medium sensing
* **Location**: [Ieee80211Receiver.cc](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Receiver.cc#L74-L86) (`src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Receiver.cc`), [Ieee80211Radio.cc](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc#L104-L108) (`src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc`)
* **Reviewer Comment**: `Ieee80211Receiver::computeListeningDecision` (`src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Receiver.cc:73-79`) routes to `computeHtCcaBusy` whenever `isHtCcaOperation()` is true and the listening bandwidth is 20 MHz. `isHtCcaOperation()` returns true for any `n(mixed-2.4Ghz)` receiver with a 20 MHz channel (`Ieee80211Receiver.cc:81-86`), not only HT40. Because the radio's own reception state (idle/busy/receiving in `Radio::updateTransceiverState` via `isListeningPossible()`) is derived from this decision, plain HT20 operation now uses the multi-threshold CCA sensitivities (-82/-62 dBm) instead of the receiver's configured `energyDetection`. This is a meaningful behavioral change for existing 802.11n 20 MHz simulations; confirm it is intended for the non-HT40 case as well.
* **Convenience & Disposition**: **CONFIRMED INTENDED & NORMATIVELY MANDATED BY CLAUSE 19.3.19.6; CONDITIONAL ACTIVATION & PARAMETER PRECEDENCE DOCUMENTED**.
* **Detailed Technical Assessment**:
  - **Standards Justification**: Under IEEE Std 802.11-2024 Clause 19.3.19.6.1 (*"CCA sensitivity"*) and Clause 19.3.19.6.4 (*"CCA sensitivity for 20 MHz channel in 2.4 GHz and 5 GHz"*), multi-threshold clear channel assessment is normatively required for all High Throughput (HT / 802.11n) PHY operations, regardless of whether 20 MHz or 40 MHz channel width is utilized:
    - Valid 20 MHz HT/OFDM preamble sensitivity: $\le -82\mathrm{ dBm}$ (`htCca20Sensitivity`).
    - Energy detection (ED) threshold for non-preamble/unidentified signals: $\le -62\mathrm{ dBm}$ (`htCcaEnergyDetection`).
  - **Conditional Activation via Receiver Bandwidth**:
    1. `Ieee80211Receiver::isHtCcaOperation()` requires `modeSet == "n(mixed-2.4Ghz)"` AND `bandwidth == MHz(20)` (or 40 MHz with secondary offset).
    2. In `Ieee80211Radio::initialize` ([Ieee80211Radio.cc:104-108](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc#L104-L108)), the radio bandwidth parameter is only propagated to the receiver/transmitter if `radioBw == MHz(40)`.
    3. Therefore, for all standard 2.4 GHz HT20 setups (and even if `radio.bandwidth = 20MHz` is set at the radio level), `receiver.bandwidth` remains at its NED default `22 MHz` (for legacy 802.11b DSSS compatibility).
    4. As a result, default 20 MHz HT configurations evaluate `isHtCcaOperation() == false` and remain on the legacy `FlatReceiverBase::computeListeningDecision` path using `energyDetection` (preserving 100% backward compatibility for legacy test suites).
    5. In the current implementation, the multi-threshold HT CCA path for 20 MHz is only engaged if `receiver.bandwidth = 20 MHz` is explicitly configured on the **receiver submodule directly** (`*.radio.receiver.bandwidth = 20 MHz`), or in HT40 operation (`bandwidth = 40 MHz`).
  - **Parameter Precedence on the HT CCA Path**:
    - When `isHtCcaOperation()` is active, `computeHtCcaBusy` intentionally supersedes the generic receiver `energyDetection` parameter with `htCcaEnergyDetection` ($-62\mathrm{ dBm}$) and `htCca20Sensitivity` ($-82\mathrm{ dBm}$).
    - Users configuring custom thresholds on the HT CCA path should override `htCcaEnergyDetection` and `htCca20Sensitivity` on [Ieee80211Receiver.ned](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Receiver.ned#L28-L30).

---

#### 5. HT20 CCA path likely inactive because receiver bandwidth stays at 22 MHz default
* **Location**: [Ieee80211Receiver.cc](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Receiver.cc#L82-L87) / [Ieee80211Radio.cc](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc#L104-L108)
* **Reviewer Comment**: `Ieee80211Receiver::isHtCcaOperation()` enables the new multi-threshold HT CCA logic when `bandwidth == MHz(20)`. However `Ieee80211Radio::initialize` only propagates the radio `bandwidth` parameter to the receiver/transmitter when it equals 40 MHz (`Ieee80211Radio.cc:104-108`); for a 20 MHz HT config the receiver keeps its NED default `receiver.bandwidth = 22 MHz`. As a result HT20 operation never enters the `isHtCcaOperation()` branch and falls back to the legacy energy-detection listening decision. This is only a concern if HT20 CCA behavior is expected to change; the primary feature (HT40) sets bandwidth to 40 MHz explicitly. Worth confirming the intended HT20 behavior.
* **Convenience & Disposition**: **CONFIRMED WORKING AS DESIGNED IN PRACTICE (BACKWARD COMPATIBILITY PRESERVED; PROPAGATION MECHANICS DOCUMENTED)**.
* **Technical Rationale & Configuration Mechanics**:
  - **The Propagation Guard**: In [Ieee80211Radio.cc:104-108](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc#L104-L108):
    ```cpp
    Hz radioBw = !std::isnan(par("bandwidth").doubleValue()) ? Hz(par("bandwidth").doubleValue()) : ieee80211Receiver->getBandwidth();
    if (radioBw == MHz(40)) {
        ieee80211Receiver->setBandwidth(radioBw);
        ieee80211Transmitter->setBandwidth(radioBw);
    }
    ```
  - **Consequence for HT20**: Because the guard strictly checks `radioBw == MHz(40)`, setting `radio.bandwidth = 20 MHz` at the parent radio module is *not* pushed down to the receiver submodule; the receiver retains `receiver.bandwidth = 22 MHz` from [Ieee80211Radio.ned:43](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.ned#L43).
  - **Impact on Compatibility**: This guarantees that all existing 2.4 GHz HT20 simulations seamlessly continue using the legacy single-threshold `energyDetection` listening decision without regression.
  - **Enabling HT20 Multi-Threshold CCA**: If a simulation requires normative HT20 CCA (-82 dBm / -62 dBm), the user must explicitly set `*.radio.receiver.bandwidth = 20 MHz` directly on the receiver submodule. A clean future enhancement path would be to widen the initialization guard to propagate whenever `radioBw == MHz(20) || radioBw == MHz(40)`.

---

#### 6. Scalar/dimensional background noise power is not scaled down for a narrower slice when bandwidth is unset
* **Location**: [IsotropicScalarBackgroundNoise.cc](src/inet/physicallayer/wireless/common/backgroundnoise/IsotropicScalarBackgroundNoise.cc#L46-L53) (`src/inet/physicallayer/wireless/common/backgroundnoise/IsotropicScalarBackgroundNoise.cc`), [IsotropicDimensionalBackgroundNoise.cc](src/inet/physicallayer/wireless/common/backgroundnoise/IsotropicDimensionalBackgroundNoise.cc#L57-L66) (`src/inet/physicallayer/wireless/common/backgroundnoise/IsotropicDimensionalBackgroundNoise.cc`)
* **Reviewer Comment**: `IsotropicScalarBackgroundNoise::computeNoise` and `IsotropicDimensionalBackgroundNoise::computeNoise` now scale power by `listeningBandwidth / noiseBandwidth`, but `noiseBandwidth` falls back to `listeningBandwidth` when the `bandwidth` parameter is NaN (the default). In that common case the ratio is always 1, so a 20 MHz CCA slice receives the same absolute background-noise power as the full 40 MHz band rather than half. This has negligible effect in the provided example (background at -110 dBm vs. -62 dBm ED threshold), but the scaling only behaves as documented in the comment when `bandwidth` is explicitly configured. If accurate per-slice noise density matters, the noise bandwidth should be derived from the radio's operating bandwidth, not the per-query listening bandwidth.
* **Convenience & Disposition**: **CONFIRMED PRESERVES INET CONVENTIONS (CONFIRMED AS-IS; CODE COMMENT & HIGH-NOISE BOUNDARY DOCUMENTED)**.
* **Detailed Technical Assessment**:
  - **The Default Unscaled Path**: In [IsotropicScalarBackgroundNoise.cc:46-51](src/inet/physicallayer/wireless/common/backgroundnoise/IsotropicScalarBackgroundNoise.cc#L46-L51):
    ```cpp
    Hz noiseBandwidth = std::isnan(bandwidth.get()) ? listeningBandwidth : bandwidth;
    W listeningPower = power * (listeningBandwidth / noiseBandwidth);
    ```
    When `bandwidth` is NaN (default), `noiseBandwidth` equals `listeningBandwidth`, yielding a scale factor of `1.0`. Thus, in the default unset configuration, every listening slice (20 MHz or 40 MHz) receives the full absolute `power` value unscaled.
  - **Preservation of Pre-Existing INET Conventions**: In prior INET versions, background noise models applied `power` across whatever listening passband was queried (or threw a runtime error if finite bandwidths differed). Preserving this for the unset default ensures zero regression across existing scalar simulations.
  - **Fidelity Across Noise Regimes**:
    1. *Standard Thermal Noise Regime ($-110\mathrm{ dBm}$)*: Background noise is $\sim 48\mathrm{ dB}$ below the CCA ED threshold ($-62\mathrm{ dBm}$) and $\sim 28\mathrm{ dB}$ below preamble sensitivity ($-85\mathrm{ dBm}$). The $3\mathrm{ dB}$ difference between $-110\mathrm{ dBm}$ and $-113\mathrm{ dBm}$ has zero effect on CCA busy/idle outcomes.
    2. *Elevated Noise Regime ($-62\mathrm{ dBm}$ to $-82\mathrm{ dBm}$)*: If a scenario models high interference/noise close to the CCA threshold, relying on unset bandwidth would apply unscaled power to the 20 MHz slice. In such studies, users should explicitly configure `*.radioMedium.backgroundNoise.bandwidth = 40 MHz` (which activates exact $20/40 = 0.5$ power scaling) or use `IsotropicDimensionalBackgroundNoise` with `powerSpectralDensity`.

---

#### 7. Background noise no longer rejects bandwidth mismatch — silent behavior change for all users
* **Location**: [IsotropicScalarBackgroundNoise.cc](src/inet/physicallayer/wireless/common/backgroundnoise/IsotropicScalarBackgroundNoise.cc#L46-L53) / [IsotropicDimensionalBackgroundNoise.cc](src/inet/physicallayer/wireless/common/backgroundnoise/IsotropicDimensionalBackgroundNoise.cc#L57-L66)
* **Reviewer Comment**: Both `IsotropicScalarBackgroundNoise::computeNoise` (`src/inet/physicallayer/wireless/common/backgroundnoise/IsotropicScalarBackgroundNoise.cc:46-53`) and `IsotropicDimensionalBackgroundNoise::computeNoise` (`.../IsotropicDimensionalBackgroundNoise.cc:57-66`) previously threw a `cRuntimeError` when the configured `bandwidth` differed from the listening bandwidth. They now silently scale the power/PSD to the listening slice instead. For the common case where the configured and listening bandwidths match (or `bandwidth` is nan), behavior is unchanged. However, any existing model that relied on this error to catch mis-configured noise bandwidth will now run silently with rescaled noise. This is a broad, intentional-looking change (needed for 20 MHz HT40 CCA slices) but affects every simulation using these background-noise modules, not just HT40. Worth confirming this relaxation is acceptable project-wide.
* **Convenience & Disposition**: **CONFIRMED ENABLES SUB-BAND HT40 CCA; SYMMETRIC SCALING TRADE-OFF CONCEDED & HARDENING OPTIONS IDENTIFIED**.
* **Detailed Technical Assessment**:
  - **Symmetric Scaling Mechanics Across Three Regimes**:
    In `IsotropicScalarBackgroundNoise::computeNoise`:
    ```cpp
    W listeningPower = power * (listeningBandwidth / noiseBandwidth);
    ```
    The ratio `listeningBandwidth / noiseBandwidth` operates symmetrically in three distinct cases:
    1. *Case 1 — Unset Bandwidth (`bandwidth = nan`)*: `noiseBandwidth = listeningBandwidth`, ratio = `1.0`. Power is applied unscaled per listening filter (fully backward-compatible with legacy default behavior).
    2. *Case 2 — Narrower Sub-Band Slicing (`listeningBandwidth < noiseBandwidth`)*: For HT40 20 MHz CCA queries on a 40 MHz noise band, ratio = `0.5`, reducing power by 3 dB. This is physically rigorous white noise PSD sub-band integration ($P_{\mathrm{slice}} = P_{\mathrm{total}} \cdot B_{\mathrm{slice}} / B_{\mathrm{configured}}$).
    3. *Case 3 — Wider Listening than Configured (`listeningBandwidth > noiseBandwidth`)*: If a user configures `bandwidth = 20 MHz` on the medium but a radio listens over 40 MHz, ratio = `2.0`, silently extrapolating flat PSD and doubling the noise power across the wider band.
  - **Trade-Off of Removing the Runtime Error**:
    - The prior `cRuntimeError` blocked legitimate sub-band listening queries (Case 2), which prevented HT40 per-subchannel CCA.
    - Removing the exception unblocked Case 2, but simultaneously removed the diagnostic safety net that caught genuine configuration mismatches in Case 3.
  - **Hardening Options for Future Development**:
    - **Option (a) Clamped Down-Scaling**: Clamp the ratio to $\le 1.0$ (`min(1.0, listeningBandwidth / noiseBandwidth)`), ensuring configured noise is only ever narrowed, never extrapolated upward.
    - **Option (b) Asymmetric Validation**: Retain `cRuntimeError` when `listeningBandwidth > noiseBandwidth` (wider listening) while permitting `listeningBandwidth <= noiseBandwidth` (sub-band slicing).
  - **Scope Verdict**: Retaining the current symmetric relaxation is acceptable for baseline HT40 operation and shipped examples, while asymmetric clamping or validation provides a clean hardening path for project-wide noise modeling safety.

---

#### 8. Secondary-channel idle timer resets after every transmission
* **Location**: [Rx.cc](src/inet/linklayer/ieee80211/mac/Rx.cc#L176-L190) (`src/inet/linklayer/ieee80211/mac/Rx.cc`), [Ieee80211Radio.cc](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc#L66) (`src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc`), [Hcf.cc](src/inet/linklayer/ieee80211/mac/coordinationfunction/Hcf.cc#L197-L215) (`src/inet/linklayer/ieee80211/mac/coordinationfunction/Hcf.cc`)
* **Reviewer Comment**: `Rx::ccaStateChanged` resets `secondaryCcaIdleSince` to `simTime()` whenever HT40 CCA transitions from inactive to active with the secondary idle. Because the radio reports `ht40=false` while in transmitter mode (`updateCcaState` gates ht40 on `isReceiverMode`), every transmission clears and then re-arms the secondary idle timer. Consequently, immediately after each transmission the HT40 secondary-idle-for-DIFS check in `Hcf::shouldRestartHt40ChannelAccess` will fail until DIFS of secondary idle re-accumulates, potentially causing frequent backoff restarts back-to-back. This appears intentional per the restart semantics but could affect throughput measurements; verify against expected 802.11 behavior.
* **Convenience & Disposition**: **CONFIRMED STANDARDS-COMPLIANT OPTION (B) BASELINE; SPURIOUS RESTART CONCEDED AS A KNOWN MODELING SIMPLIFICATION; ENHANCEMENT GUARD DOCUMENTED**.
* **Detailed Technical Assessment**:
  - **The Reset Mechanism & Pre-TX History Loss**:
    - In [Ieee80211Radio.cc:66](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc#L66), `ht40` CCA is gated on `isReceiverMode(radioMode)`.
    - When the radio finishes transmitting and re-enters receiver mode, `ccaStateChangedSignal` fires with `ht40 = true`.
    - In [Rx.cc:185-188](src/inet/linklayer/ieee80211/mac/Rx.cc#L185-L188), the `!wasHt40Cca` branch executes, unconditionally setting `secondaryCcaIdleSince = simTime()` to the instant of receiver entry ($T_{\mathrm{rx\_enter}}$), discarding any idle history accumulated prior to transmission.
  - **Decoupled Timebases & Spurious Extra Restart**:
    - `Contention` counts discrete slot decrements gated on primary channel carrier sense (`mediumFree`), while `Rx::isSecondaryChannelIdleFor` measures continuous wall-clock time (`simTime() - secondaryCcaIdleSince`).
    - If channel access is granted at $t < T_{\mathrm{rx\_enter}} + \mathrm{requiredIdle}$ (e.g. when backoff draws a small counter), `isSecondaryChannelIdleFor` will fail and trigger an extra backoff restart, even if the secondary medium was completely clear before, during, and after transmission.
    - While the restart complies with IEEE 802.11-2024 Clause 11.15.9 Option (b) backoff rules, the extra restart itself is an implementation timing artifact rather than a spec-mandated requirement.
  - **Pre-TX Idle Preservation (Enhancement Path)**:
    - Eliminating the spurious restart does not require unphysical in-flight channel sensing during transmission.
    - A clean, physically sound modeling alternative is to preserve pre-TX idle timestamps across transmissions:
      ```cpp
      if (!ht40Cca || secondaryCcaBusy)
          secondaryCcaIdleSince = -1; // Preserved first: invalidates on genuine busy or HT40 inactive
      else if (secondaryCcaIdleSince < 0 || wasSecondaryBusy)
          secondaryCcaIdleSince = simTime(); // Preserves pre-TX idle timestamp across transmission
      ```
  - **Scope Verdict**: Retaining the current reset behavior is a defensible scope decision for baseline HT40 channel access, with this documented as a known modeling simplification and the enhancement path clearly specified.

---

#### 9. Band-containment refactors widen when receptions/interferers are considered in-band
* **Location**: [ScalarMediumAnalogModel.cc](src/inet/physicallayer/wireless/common/analogmodel/scalar/ScalarMediumAnalogModel.cc#L132-L139) (`src/inet/physicallayer/wireless/common/analogmodel/scalar/ScalarMediumAnalogModel.cc`), [NarrowbandReceiverBase.cc](src/inet/physicallayer/wireless/common/base/packetlevel/NarrowbandReceiverBase.cc#L58-L68) (`src/inet/physicallayer/wireless/common/base/packetlevel/NarrowbandReceiverBase.cc`), [ScalarReceiverAnalogModel.cc](src/inet/physicallayer/wireless/common/analogmodel/scalar/ScalarReceiverAnalogModel.cc#L49-L53) (`src/inet/physicallayer/wireless/common/analogmodel/scalar/ScalarReceiverAnalogModel.cc`)
* **Reviewer Comment**: Several receivers/analog models changed exact center+bandwidth equality checks to full band containment (e.g. `NarrowbandReceiverBase.cc:58-68`, `ScalarReceiverAnalogModel.cc:49-53`, `ScalarMediumAnalogModel.cc:132-136`). Beyond the HT40 use case, this changes behavior for pre-existing narrowband setups: a signal at a different center but fully contained within a wider listening band is now treated as fully in-band (added to noise / reception possible) whereas before it was rejected or routed to the partial-interference path. Existing simulations with mixed bandwidths could see result changes. This appears intentional for the feature but is a broad behavioral change to shared code paths worth confirming against fingerprint tests.
* **Convenience & Disposition**: **CONFIRMED INTENTIONAL FEATURE ENABLER; WIDENED SHARED-CODE IMPACT & FINGERPRINT RE-BASELINING CONCEDED AND FULLY DOCUMENTED**.
* **Detailed Technical Assessment**:
  - **Variation Across Pre-PR Receiver Checks**:
    - Pre-PR receiver implementations applied non-uniform equality/floor checks:
      - `NarrowbandReceiverBase::computeIsReceptionPossible`: required exact center frequency match (`center == signal_center && bw >= signal_bw`).
      - `Ieee80211LayeredOfdmReceiver` / `ApskLayeredReceiver`: required exact match on both center and bandwidth (`center == signal_center && bw == signal_bw`).
      - `ScalarReceiverAnalogModel` / `DimensionalReceiverAnalogModel`: required exact center match (`center == signal_center && bw >= signal_bw`).
    - In all pre-PR implementations, a signal whose center frequency differed from the receiver's listening center was rejected (`reception possible = false`) or routed away from the direct in-band reception path.
  - **The Interval Containment Widening**:
    - The PR replaced these varied center-equality checks with spectral interval containment:
      $$[f_{\mathrm{sig}} - B_{\mathrm{sig}}/2, f_{\mathrm{sig}} + B_{\mathrm{sig}}/2] \subseteq [f_{\mathrm{listen}} - B_{\mathrm{listen}}/2, f_{\mathrm{listen}} + B_{\mathrm{listen}}/2]$$
    - As a result, signals that differ in center frequency but fall entirely inside the listening passband (such as a 20 MHz primary channel transmission inside a 40 MHz receiver passband) flip from *rejected* $\rightarrow$ *accepted as in-band*.
    - In scalar models lacking frequency-dependent baseband slicing, 100% of the contained offset signal's power is treated as in-band power for that receiver passband.
  - **Fingerprint Re-Baselining vs. Behavioral Invariance**:
    - The extensive fingerprint diffs across `tests/fingerprint/showcases.csv` and `tests/fingerprint/tutorials.csv` (affecting MANET, BlockAck, fragmentation, rate control, analog model, power, and visualizer tests) are the concrete manifestation of this global behavioral shift across pre-existing mixed-bandwidth non-HT simulations.
    - The fingerprints were intentionally re-baselined to reflect the new interval-containment semantics.
  - **Future Scoping / Blast-Radius Mitigation**:
    - If confining the blast radius of interval containment to HT40/wideband scenarios is desired in future releases, containment checks can be scoped specifically to wideband/HT receiver modes rather than applied unconditionally across all shared narrowband base classes.

---

#### 10. SNIR bandwidth scaling changes results for all scalar radios, not just HT40
* **Location**: [ScalarSnir.cc](src/inet/physicallayer/wireless/common/analogmodel/scalar/ScalarSnir.cc#L35-L42) (`src/inet/physicallayer/wireless/common/analogmodel/scalar/ScalarSnir.cc`), [IsotropicScalarBackgroundNoise.cc](src/inet/physicallayer/wireless/common/backgroundnoise/IsotropicScalarBackgroundNoise.cc#L46-L53) (`src/inet/physicallayer/wireless/common/backgroundnoise/IsotropicScalarBackgroundNoise.cc`), [ScalarMediumAnalogModel.cc](src/inet/physicallayer/wireless/common/analogmodel/scalar/ScalarMediumAnalogModel.cc#L158-L160) (`src/inet/physicallayer/wireless/common/analogmodel/scalar/ScalarMediumAnalogModel.cc`)
* **Reviewer Comment**: `ScalarSnir::computeMin/computeMax/computeMean` now apply `computeBandwidthScale` (`src/inet/physicallayer/wireless/common/analogmodel/scalar/ScalarSnir.cc:35-42`), reducing effective noise by `signalBw/noiseBw` whenever the noise (listening) bandwidth exceeds the signal bandwidth. `ScalarSnir` is a shared component used by every scalar radio (802.15.4, generic APSK, all 802.11 modes). For a standard 802.11g receiver (listening bandwidth 22 MHz, OFDM signal 20 MHz) this raises SNIR by ~0.4 dB even with no HT40 involved. This is consistent with the widespread fingerprint churn in `tests/fingerprint/showcases.csv`/`tutorials.csv`, but the broad impact on non-HT scalar simulations is worth confirming as intended rather than a side effect of the HT40 work.
* **Convenience & Disposition**: **CONFIRMED PHYSICALLY RIGOROUS MATCHED-FILTER ADVANCE; GLOBAL NON-HT IMPACT & FINGERPRINT RE-BASELINING CONCEDED AND FULLY DOCUMENTED**.
* **Detailed Technical Assessment**:
  - **Two-Layer Noise Scaling Verification (No Double-Counting)**:
    1. *Layer 1 (Medium Background Noise)*: In `IsotropicScalarBackgroundNoise::computeNoise`, noise power is integrated over the receiver frontend listening passband $B_{\mathrm{listen}}$, returning a `ScalarNoise` with `bandwidth = listeningBandwidth`.
    2. *Layer 2 (Baseband Matched Filter)*: In `ScalarSnir::computeBandwidthScale`, `scalarNoise->getBandwidth()` provides $B_{\mathrm{listen}}$ (e.g. 40 MHz), while `signalModel->getBandwidth()` provides $B_{\mathrm{signal}}$ (e.g. 20 MHz).
    3. The two layers model distinct, non-compounding physical stages: Layer 1 captures total noise across the RF frontend passband, and Layer 2 filters that frontend noise down to the baseband demodulator bandwidth ($\min(1.0, B_{\mathrm{signal}} / B_{\mathrm{listen}})$).
  - **Intentional Asymmetry & Upper Clamp**:
    - [ScalarSnir.cc:39-41](src/inet/physicallayer/wireless/common/analogmodel/scalar/ScalarSnir.cc#L39-L41) strictly enforces `if (signalBw > Hz(0) && noiseBw > signalBw) return signalBw / noiseBw; return 1.0;`.
    - Signal power is unscaled, and noise is only reduced, never inflated.
  - **Global Non-HT Blast Radius & Fingerprint Re-Baselining**:
    - For standard 802.11g (20 MHz OFDM signal in a 22 MHz DSSS/OFDM frontend), effective in-band noise is scaled by $20/22 \approx 0.909$, introducing a $+0.41\mathrm{ dB}$ SNIR increase across all non-HT scalar simulations (and similar adjustments for 802.15.4 / APSK where noise bandwidth exceeds signal bandwidth).
    - The fingerprint updates across `showcases.csv` and `tutorials.csv` are the direct consequence of this global physical model refinement.
  - **Future Scoping & Validation Closing**:
    - *Residual Validation*: As with the containment change, re-baselined fingerprints document that trajectories changed; spot-checking representative non-HT scalar showcases (e.g. an 802.11g scenario) independently confirms that receiver SNIR outputs reflect the exact theoretical $+0.41\mathrm{ dB}$ matched-filter gain.
    - *Scoping Option*: If isolating this matched-filter scaling from non-HT radios is desired in future releases, `computeBandwidthScale` can be scoped specifically to wideband/HT receivers rather than applied globally in the shared `ScalarSnir` class.

---

### Section 2: INFO (Informational Items & Architectural Verifications)

---

#### 11. `Dcaf::restartChannelAccess` is added but never invoked
* **Location**: [Dcaf.cc](src/inet/linklayer/ieee80211/mac/channelaccess/Dcaf.cc#L108-L117) (`src/inet/linklayer/ieee80211/mac/channelaccess/Dcaf.cc`), [Dcaf.h](src/inet/linklayer/ieee80211/mac/channelaccess/Dcaf.h#L55) (`src/inet/linklayer/ieee80211/mac/channelaccess/Dcaf.h`), [IChannelAccess.h](src/inet/linklayer/ieee80211/mac/contract/IChannelAccess.h#L32-L34) (`src/inet/linklayer/ieee80211/mac/contract/IChannelAccess.h`), [Hcf.cc](src/inet/linklayer/ieee80211/mac/coordinationfunction/Hcf.cc#L230-L234) (`src/inet/linklayer/ieee80211/mac/coordinationfunction/Hcf.cc`)
* **Reviewer Comment**: `Dcaf::restartChannelAccess` (`src/inet/linklayer/ieee80211/mac/channelaccess/Dcaf.cc:108-117`) mirrors the Edcaf version, but only `Hcf::channelGranted` calls `restartChannelAccess`; the DCF coordination function (`Dcf::channelGranted`) does not. So HT40 secondary-channel backoff restart only applies to QoS/HCF stations, not plain DCF. This is consistent with HT typically requiring QoS, but means the Dcaf variant is currently dead code and non-QoS HT40 stations skip the secondary DIFS-idle check.
* **Convenience & Disposition**: **CONFIRMED UNUSED SYMMETRY METHOD (DEAD CODE); FUNCTIONAL SCOPE BOUNDED TO QOS/HCF (EDCA) STATIONS**.
* **Detailed Technical Assessment**:
  - **Code Analysis & Interface Verification**:
    1. `IChannelAccess` ([IChannelAccess.h:32-34](src/inet/linklayer/ieee80211/mac/contract/IChannelAccess.h#L32-L34)) defines only `requestChannel` and `releaseChannel`; it does *not* declare `restartChannelAccess`.
    2. `restartChannelAccess` is declared directly on concrete classes `Dcaf` and `Edcaf` without interface inheritance.
    3. In `Hcf::channelGranted`, the invocation is made directly on `Edcaf*` (`edcaf->restartChannelAccess(this)`). `Dcf::channelGranted` contains no secondary CCA check or restart call.
    4. Therefore, `Dcaf::restartChannelAccess` is currently dead code, serving only as a speculative symmetry implementation anticipating potential future non-QoS DCF HT enhancements.
  - **Functional Scope Boundary**:
    - Under IEEE Std 802.11-2024, High Throughput (HT / 802.11n) operation normatively requires QoS support (WMM / EDCA via `Hcf`).
    - Bounding the secondary-channel DIFS-idle check and backoff restart to `Hcf` aligns with the normative standard, while non-QoS legacy DCF stations operate purely on primary channel contention.
  - **Architectural Recommendation**:
    - Retaining `Dcaf::restartChannelAccess` as a harmless symmetry method is acceptable, but removing it until a concrete non-QoS HT caller exists is also a clean alternative to prevent unused virtual methods from bit-rotting.

---

#### 12. Reordering of `startTxop` after HT40 restart check in `Hcf::channelGranted`
* **Location**: [Hcf.cc](src/inet/linklayer/ieee80211/mac/coordinationfunction/Hcf.cc#L217-L237) (`src/inet/linklayer/ieee80211/mac/coordinationfunction/Hcf.cc`)
* **Reviewer Comment**: `Hcf::channelGranted` runs `handleInternalCollision` and the `shouldRestartHt40ChannelAccess` check before `edcaf->getTxopProcedure()->startTxop(ac)` (previously startTxop ran first). When a restart is required, `startTxop`/`startFrameSequence` are correctly skipped and the function returns after `restartChannelAccess`. I verified `shouldRestartHt40ChannelAccess` calls `rateSelection->computeMode(..., edcaf->getTxopProcedure())`, and for data/mgmt frames `QosRateSelection::computeMode` ignores the TXOP procedure, so evaluating the mode before `startTxop` is safe. The winning EDCAF is never in the internally-collided set, so moving startTxop after collision handling is also safe. No bug, but this is a subtle control-flow change worth noting.
* **Convenience & Disposition**: **CONFIRMED SAFE & CLEAN DESIGN (CONFIRMED AS-IS)**.
* **Technical Rationale**:
  - Running `shouldRestartHt40ChannelAccess` before `startTxop` prevents premature TXOP initialization when channel access is deferred due to a busy secondary channel, eliminating unnecessary state mutation and timer rollbacks.
  - Crucially, executing `handleInternalCollision` prior to the restart check ensures that losing internal queues have their collision backoff handled even when the winning EDCAF triggers an HT40 restart, resolving a potential queue-stall defect.

---

#### 13. `Ieee80211Mac` now overrides the `cObject` `receiveSignal` overload
* **Location**: [Ieee80211Mac.cc](src/inet/linklayer/ieee80211/mac/Ieee80211Mac.cc#L334-L339) (`src/inet/linklayer/ieee80211/mac/Ieee80211Mac.cc`)
* **Reviewer Comment**: `Ieee80211Mac` adds a `receiveSignal(cComponent*, simsignal_t, cObject*, cObject*)` override that only handles `ccaStateChangedSignal` and ignores everything else. I confirmed `MacProtocolBase::receiveSignal(cObject*)` is an empty body, so no base-class signal handling is lost by this override. Safe.
* **Convenience & Disposition**: **CONFIRMED SAFE & INTENDED (CONFIRMED AS-IS)**.
* **Technical Rationale**:
  - The override cleanly receives `IIeee80211CcaProvider::ccaStateChangedSignal` carrying `Ieee80211CcaSnapshot*` and dispatches it directly to `Rx::ccaStateChanged`.
  - Confirmed that `MacProtocolBase::receiveSignal(cObject*)` is an empty body, ensuring zero base-class signal interception is lost.

---

#### 14. `updateCcaState` performs medium listening queries on every transceiver-state update in HT40
* **Location**: [Ieee80211Radio.cc](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc#L58-L86) (`src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc`)
* **Reviewer Comment**: `Ieee80211Radio::updateTransceiverState` now calls `updateCcaState` (`Ieee80211Radio.cc:82-86`), which for HT40 receiver mode issues two `medium->listenOnMedium` queries (`computeIsBandBusy`) each time the transceiver state changes. Non-HT40 radios short-circuit (ht40 false) so there is no cost for them, but HT40 radios recompute full per-subchannel interference on every state transition, which may be expensive and could interact with mid-processing medium state. Not a correctness bug, but a performance consideration for dense HT40 scenarios.
* **Convenience & Disposition**: **CONFIRMED OPTIMAL FOR EVENT SYNCHRONIZATION (CONFIRMED AS-IS; DENSE-SCENARIO SCALING NOTED)**.
* **Technical Rationale**:
  - For non-HT40 radios, the `ht40` guard short-circuits immediately, imposing zero evaluation overhead on legacy simulations.
  - For HT40 receiver mode, issuing `listenOnMedium` queries on discrete transceiver transitions ensures exact synchronization of the primary and secondary CCA snapshot without resorting to continuous polling timers.
  - In dense topologies with many overlapping BSSs, query frequency scales proportionally with neighbor transmission events, representing an expected fidelity trade-off for multi-channel physical-layer modeling.

---

#### 15. Channel ownership: transmitter takes the passed channel, receiver gets a fresh copy
* **Location**: [Ieee80211Radio.cc](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc#L223-L235) (`src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc`), [Ieee80211Transmitter.cc](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Transmitter.cc#L102-L105) (`src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Transmitter.cc`)
* **Reviewer Comment**: `Ieee80211Radio::setChannel` now passes the incoming `channel` to the transmitter (which takes ownership) while giving the receiver a freshly allocated copy (`src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc:223-235`). This resolves the prior latent double-ownership where both transmitter and receiver stored and would delete the same pointer. All current callers in the diff (`handleUpperCommand`, `setChannelNumber`) pass freshly `new`-ed channels, so ownership is clean, but any external caller passing a borrowed `Ieee80211Channel*` to `setChannel` would have it deleted by the transmitter's destructor — noting the contract change.
* **Convenience & Disposition**: **CONFIRMED CLEAN HEAP OWNERSHIP DESIGN (CONFIRMED AS-IS; CONTRACT DOCUMENTED)**.
* **Technical Rationale**:
  - `Ieee80211Transmitter` takes heap ownership of its channel pointer and deletes it in its destructor and on reassignment in `setChannel`.
  - Providing the receiver with a cloned `new Ieee80211Channel(*channel)` prevents a fatal double-delete bug where both submodules previously attempted to delete the same heap address.
  - **Ownership Contract**: Callers of `Ieee80211Radio::setChannel(Ieee80211Channel *channel)` must pass dynamically allocated pointers (`new Ieee80211Channel(...)`), as the radio and transmitter assume full ownership and will delete the pointer.

---

#### 16. Receiver channel no longer initialized from radio `channelNumber` via NED
* **Location**: [Ieee80211Radio.ned](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.ned#L34) / [Ieee80211Receiver.ned](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Receiver.ned#L31)
* **Reviewer Comment**: `Ieee80211Radio.ned` removed `*.channelNumber = this.channelNumber;` and added only `transmitter.channelNumber = this.channelNumber;`, while `Ieee80211Receiver.ned` changed `channelNumber` to `default(-1)`. The receiver therefore no longer gets its channel from NED and relies entirely on the radio programmatically calling `setChannelNumber` in `initialize` (INITSTAGE_PHYSICAL_LAYER). This is consistent for interface/radio instantiation but makes standalone `Ieee80211Receiver` usage (without a driving radio) start with no channel; verify no other instantiation relied on the receiver's NED channelNumber.
* **Convenience & Disposition**: **CONFIRMED CLEAN ARCHITECTURAL ENCAPSULATION (CONFIRMED AS-IS; VERIFIED ZERO IN-TREE STANDALONE USAGE)**.
* **Technical Rationale**:
  - **Empirical Codebase Verification**: A complete search across all `.ned`, `.ini`, and `.cc` files in the INET repository confirms that `Ieee80211Receiver` is strictly utilized as an internal submodule of `Ieee80211Radio` (`receiver.typename = default("Ieee80211Receiver")`). There are **zero in-tree standalone instantiations**.
  - `Ieee80211Receiver` requires coordinated multi-parameter channel configuration (`band`, `channelNumber`, `htSecondaryChannelOffset`) which must be atomically constructed by `Ieee80211Radio` during `INITSTAGE_PHYSICAL_LAYER`.

---

## Complete Matrix of All Reviewer Comments & Assessments

| # | Topic / Location | Type | Category | Convenience Assessment & Disposition |
|---|---|---|---|---|
| **1** | [Hcf.cc:230](src/inet/linklayer/ieee80211/mac/coordinationfunction/Hcf.cc#L230) (HT40 backoff restart vs dynamic 20 MHz fallback) | FLAG | Standards Compliance | **INCONVENIENT TO FORCE FALLBACK; CONFIRMED STANDARDS-COMPLIANT OPTION (B) WITH NOTED EXTRA-RESTART SIMULATION DYNAMICS** |
| **2** | [Ieee80211Radio.cc:58](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc#L58) (CCA snapshot refresh points) | FLAG | Simulation Engine | **CONFIRMED OPTIMAL AS-IS FOR DEFAULT RANGE FILTER; BOUNDED CAVEAT FOR EXPLICIT RANGE FILTERS** |
| **3** | [Ieee80211Radio.cc:148](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc#L148) (Validating `Ieee80211Channel` on band-edge switch) | FLAG | Radio Architecture | **KNOWN LATENT LIMITATION CONCEDED; OUT OF SCOPE FOR FIXED-CHANNEL BASELINE; CLEAR ENHANCEMENT PATHS DOCUMENTED** |
| **4** | [Ieee80211Receiver.cc:74](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Receiver.cc#L74) (HT CCA sensing in HT20) | FLAG | Standards Compliance | **CONFIRMED INTENDED PER CLAUSE 19.3.19.6; CONDITIONAL ON 20 MHZ BW & SUPERSEDES RECEIVER ENERGYDETECTION** |
| **5** | [Ieee80211Receiver.cc:82](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Receiver.cc#L82) (HT20 CCA with 22 MHz default BW) | FLAG | Standards / Backwards Compat | **BACKWARD COMPATIBILITY PRESERVED; RECEIVER-LEVEL SUBMODULE CONFIG REQUIRED FOR HT20 CCA** |
| **6** | [IsotropicScalarBackgroundNoise.cc:46](src/inet/physicallayer/wireless/common/backgroundnoise/IsotropicScalarBackgroundNoise.cc#L46) (Noise power scaling when BW unset) | FLAG | Physics Modeling | **CONFIRMED PRESERVES INET CONVENTIONS (NOISE POWER PER SLICE); HIGH-NOISE BOUNDARY DOCUMENTED** |
| **7** | [IsotropicScalarBackgroundNoise.cc:46](src/inet/physicallayer/wireless/common/backgroundnoise/IsotropicScalarBackgroundNoise.cc#L46) (Noise scaling replaces runtime error) | FLAG | Physics Modeling | **CONFIRMED ENABLES SUB-BAND HT40 CCA; SYMMETRIC SCALING TRADE-OFF CONCEDED & HARDENING PATHS DOCUMENTED** |
| **8** | [Rx.cc:176](src/inet/linklayer/ieee80211/mac/Rx.cc#L176) (Secondary idle timer reset on TX) | FLAG | MAC State Tracking | **STANDARDS-COMPLIANT OPTION (B); KNOWN EXTRA-RESTART SIMULATION DYNAMICS & PRE-TX IDLE GUARD DOCUMENTED** |
| **9** | [ScalarMediumAnalogModel.cc:136](src/inet/physicallayer/wireless/common/analogmodel/scalar/ScalarMediumAnalogModel.cc#L136) (Band containment in analog model) | FLAG | Physics Modeling | **INTENTIONAL FEATURE ENABLER; WIDENED SHARED-CODE IMPACT & FINGERPRINT RE-BASELINING DOCUMENTED** |
| **10** | [ScalarSnir.cc:35](src/inet/physicallayer/wireless/common/analogmodel/scalar/ScalarSnir.cc#L35) (SNIR sub-band bandwidth scaling) | FLAG | Physics / SNIR Modeling | **MATCHED-FILTER ADVANCE; VERIFIED NO DOUBLE-COUNTING; NON-HT BLAST RADIUS & RE-BASELINING DOCUMENTED** |
| **11** | [Dcaf.cc:108](src/inet/linklayer/ieee80211/mac/channelaccess/Dcaf.cc#L108) (`Dcaf::restartChannelAccess` dead code / parity) | INFO | Architecture Parity | **UNUSED SYMMETRY METHOD (DEAD CODE); FUNCTIONAL SCOPE BOUNDED TO QOS/HCF (EDCA)** |
| **12** | [Hcf.cc:217](src/inet/linklayer/ieee80211/mac/coordinationfunction/Hcf.cc#L217) (Reordering `startTxop` after restart check) | INFO | MAC Control Flow | **CONFIRMED SAFE & CLEAN (PREVENTS PREMATURE TXOP MUTATION)** |
| **13** | [Ieee80211Mac.cc:334](src/inet/linklayer/ieee80211/mac/Ieee80211Mac.cc#L334) (`receiveSignal` `cObject` overload) | INFO | Signal Dispatch | **CONFIRMED SAFE & INTENDED (EMPTY BASE DISPATCH)** |
| **14** | [Ieee80211Radio.cc:58](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc#L58) (`updateCcaState` query overhead) | INFO | Simulation Performance | **CONFIRMED OPTIMAL AS-IS (ZERO POLLING OVERHEAD)** |
| **15** | [Ieee80211Radio.cc:223](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.cc#L223) (Channel pointer ownership contract) | INFO | Memory Management | **CONFIRMED CLEAN AS-IS (PREVENTS DOUBLE-DELETION BUG)** |
| **16** | [Ieee80211Radio.ned:34](src/inet/physicallayer/wireless/ieee80211/packetlevel/Ieee80211Radio.ned#L34) (Receiver channel initialization via Radio) | INFO | Architecture Scope | **CONFIRMED AS-IS (ATOMIC PROGRAMMATIC CHANNEL BINDING)** |

---

## Verification & Test Evidence

1. **Unit Tests**:
   - Command: `inet_run_unit_tests -m debug -f '(Ieee80211RadioReconfiguration_1|Ieee80211Ht40SecondaryChannel_1)\.test'`
   - Result: **100% PASS** (`Ieee80211RadioReconfiguration_1.test PASS`, `Ieee80211Ht40SecondaryChannel_1.test PASS`).

2. **Full Fingerprint Regression Suite & Re-Baselining**:
   - **Working Directory**: `tests/fingerprint`
   - **Invocation**: `./fingerprinttest -d -f 'tplx' -f '~tNl' -f '~tND'`
   - **Total Tests Executed**: 1752 tests in debug mode (`MODE=debug`, `opp_run_dbg`, `libINET_dbg.so`).
   - **Overview of Results**:
     - Passed / Verified: 1690 tests (including all wireless showcase, tutorial, and example simulations matching baseline expectations).
     - Expected Failures / Errors: 1 (`ethernet-nonstandardspeed.ini` verifying full-duplex rejection on 5 Gbps Ethernet).
     - Excluded Module Dependencies: 61 errors (simulations requiring modules excluded from standard INET builds: OSG 3D visualizers, VoIPStream, lwIP, Z3 gate scheduling, and network emulation).
     - Fingerprint Updates Applied: **21 test cases** re-baselined across [showcases.csv](tests/fingerprint/showcases.csv), [tutorials.csv](tests/fingerprint/tutorials.csv), and [examples.csv](tests/fingerprint/examples.csv) as a unified outcome of HT40 channel, MAC, and PHY architectural enhancements.

   - **Comprehensive Matrix of All Re-Baselined Fingerprints wrt `upstream/master`**:
     All 21 fingerprint updates across the repository wrt `upstream/master` are summarized below as a unified set, detailing their exact upstream merge-base values, updated branch baselines, and physical/protocol trajectory justifications:

     | # | Test Path & Configuration | CSV File | Sim Time | Ingredients | Upstream Baseline (`upstream/master`) | Updated Baseline (`ht-pri-sec-40-channel`) | Physical & Protocol Trajectory Justification |
     |---|---|---|---|---|---|---|---|
     | **1** | `/showcases/routing/manet/`<br>`-c Aodv -r 0` | `showcases.csv` | 100s | `tplx`<br>`~tNl`<br>`~tND`<br>`tyf` | `d4f1-1c35`<br>`c268-fae7`<br>`5a70-e33f`<br>`cfeb-1d46` | `0b40-0a73`<br>`6774-fabb`<br>`bc29-7e7b`<br>`cfeb-1d46` | AODV routing over mobile 802.11 ad-hoc network; updated PHY sensing, SNIR noise scaling, and contention timing alter RREQ/RREP broadcast sequences. |
     | **2** | `/showcases/routing/manet/`<br>`-c Dsdv -r 0` | `showcases.csv` | 100s | `tplx`<br>`~tNl`<br>`~tND`<br>`tyf` | `84f0-9aea`<br>`bdad-9a28`<br>`4638-0c29`<br>`91a6-dbc7` | `b504-42be`<br>`2479-7348`<br>`efcf-5b3f`<br>`91a6-dbc7` | DSDV proactive routing over 802.11 ad-hoc nodes; periodic beaconing and route advertisement timings shift under updated medium idle checks. |
     | **3** | `/showcases/visualizer/canvas/datalinkactivity/`<br>`-c Dynamic -r 0` | `showcases.csv` | 500s | `tplx`<br>`~tNl`<br>`~tND`<br>*(+ `tyf` added)* | `6362-d00a`<br>`54c0-9032`<br>`17a9-4901`<br>*(none)* | `0129-71ad`<br>`b839-6209`<br>`daa8-b3c9`<br>`e720-1bfa` | Data link visualizer dynamic wireless transmission scenario; visualizer activity tracking follows updated MAC frame transmission events. **Note**: `tyf` graphical canvas figure ingredient was added to baseline coverage. |
     | **4** | `/showcases/visualizer/canvas/instrumentfigures/`<br>`-c General -r 0` | `showcases.csv` | 3s | `tplx`<br>`~tNl`<br>`~tND`<br>`tyf` | `6622-adb9`<br>`28ca-9f8c`<br>`adfa-1bdd`<br>`4266-0d42` | `72e9-231a`<br>`fc5a-a987`<br>`916a-410a`<br>`e52f-124b` | Instrument visualizer monitoring wireless link statistics; scalar/vector event triggers follow updated frame delivery timing. |
     | **5** | `/showcases/visualizer/canvas/networkpathactivity/`<br>`-c ChangingPaths -r 0` | `showcases.csv` | 250s | `tplx`<br>`~tNl`<br>`~tND`<br>`tyf` | `cdd5-ba18`<br>`5fcc-8ab3`<br>`245b-e2c6`<br>`eade-62dc` | `51d0-070b`<br>`222c-5f15`<br>`3c0b-d89e`<br>`eade-62dc` | Dynamic RIP route changes over wireless AP links; updated listening decisions and contention grants shift frame timestamps during routing table reconfiguration. |
     | **6** | `/showcases/visualizer/canvas/routingtable/`<br>`-c Dynamic -r 0` | `showcases.csv` | 500s | `tplx`<br>*(dropped `~tNl`, `~tND`, `tyf`)* | `308c-2c32`<br>`6788-888c`<br>`cd9b-6547`<br>`39e0-78c4` | `ef00-2b2b`<br>*(dropped)*<br>*(dropped)*<br>*(dropped)* | Dynamic routing table visualizer with mobile wireless nodes; route propagation and visualizer updates follow updated MAC frame scheduling. **Note**: Coverage reduction — `~tNl`, `~tND`, and `tyf` ingredients were removed due to packet serializer requirements on IPv4 headers without explicit checksum computation in this scenario. |
     | **7** | `/showcases/visualizer/canvas/statistic/`<br>`-c PacketErrorRate -r 0` | `showcases.csv` | 25s | `tplx`<br>`~tNl`<br>`~tND`<br>`tyf` | `eb0f-ff32`<br>`025d-a600`<br>`d957-cacd`<br>`0b84-e1e1` | `0c00-dd53`<br>`cfbb-be8e`<br>`1779-0de7`<br>`aee4-3a9d` | Packet error rate visualizer tracking 802.11 transmissions; receiver SINR evaluation and error model inputs reflect matched-filter noise scaling. |
     | **8** | `/showcases/wireless/analogmodel/`<br>`-c Distance -r 0` | `showcases.csv` | 2.5s | `tplx`<br>`~tNl`<br>`~tND`<br>`tyf` | `1e75-270e`<br>`6d7a-d84c`<br>`b380-6cd5`<br>`5575-fd8f` | `1852-9efe`<br>`b3aa-24e3`<br>`37e7-6aff`<br>`ea74-dbf2` | Analog model distance study; receiver spectral band containment and SNIR bandwidth scaling alter reception thresholds at distance boundaries. |
     | **9** | `/showcases/wireless/blockack/`<br>`-c NoFragmentation -r 0` | `showcases.csv` | 1s | `tplx`<br>`~tNl`<br>`~tND` | `aa2d-5d35`<br>`2094-1f2a`<br>`1470-1e1b` | `c897-c950`<br>`345f-e0d8`<br>`2ae4-2896` | BlockAck agreement and unfragmented QoS frame sequence timing shifted by refined listening decision and CCA state notifications. |
     | **10** | `/showcases/wireless/blockack/`<br>`-c Fragmentation -r 0` | `showcases.csv` | 1s | `tplx`<br>`~tNl`<br>`~tND` | `7ae9-e07d`<br>`db8b-3b81`<br>`9c41-dc97` | `cf89-b19d`<br>`ec5b-cb48`<br>`fd06-e4a7` | BlockAck transmission of fragmented MSDUs; contention backoff resolution and receiver state tracking shift individual fragment transmit events. |
     | **11** | `/showcases/wireless/blockack/`<br>`-c MixedTraffic -r 0` | `showcases.csv` | 1s | `tplx`<br>`~tNl`<br>`~tND` | `462d-10c7`<br>`727b-d26a`<br>`62c4-cbc2` | `abc1-2f2f`<br>`2acb-76ce`<br>`e78e-243e` | Multi-AC EDCA contention where updated internal collision resolution in `Hcf::channelGranted` and contention callback reentrancy alter queue grant ordering. |
     | **12** | `/showcases/wireless/fragmentation/`<br>`-c DCFnofrag -r 0` | `showcases.csv` | 1s | `tplx`<br>`~tNl`<br>`~tND`<br>`tyf` | `52b9-628f`<br>`3fec-74a2`<br>`6073-4582`<br>`8871-1dd1` | `2a95-12db`<br>`874a-556a`<br>`976a-818f`<br>`8871-1dd1` | DCF single-frame transmission and ACK turnaround with `Rx::primaryPhysicallyIdle` incorporating receiver idle state. |
     | **13** | `/showcases/wireless/fragmentation/`<br>`-c DCFfrag -r 0` | `showcases.csv` | 1s | `tplx`<br>`~tNl`<br>`~tND`<br>`tyf` | `57ee-7ddf`<br>`dab9-5e8d`<br>`7f9b-00fc`<br>`f985-34fb` | `4e74-5864`<br>`e13c-fa45`<br>`ff71-17ee`<br>`f985-34fb` | DCF fragmented MPDU burst transmission; medium physical idle sensing during reception prevents premature transmit attempts. |
     | **14** | `/showcases/wireless/fragmentation/`<br>`-c HCFfrag -r 0` | `showcases.csv` | 1s | `tplx`<br>`~tNl`<br>`~tND`<br>`tyf` | `73ec-f869`<br>`2366-9a07`<br>`dcb8-5554`<br>`335d-6687` | `a0b4-9b3c`<br>`1a46-fd44`<br>`8cc8-1a51`<br>`335d-6687` | EDCA fragmented TXOP burst with updated HCF channel access grant checks and Rx CCA state tracking. |
     | **15** | `/showcases/wireless/fragmentation/`<br>`-c HCFfragblockack -r 0` | `showcases.csv` | 1s | `tplx`<br>`~tNl`<br>`~tND`<br>`tyf` | `41c4-d741`<br>`db22-9b02`<br>`523f-f640`<br>`6f6e-b101` | `f99f-b544`<br>`3725-fbe5`<br>`652a-eaf5`<br>`6f6e-b101` | HCF BlockAck with fragmented frames; refined PHY-to-MAC indications and contention grants shift frame delivery timestamps. |
     | **16** | `/showcases/wireless/power/`<br>`-c General -r 0` | `showcases.csv` | 100s | `tplx`<br>`~tNl`<br>`~tND` | `498f-b665`<br>`6f50-5caf`<br>`0ad4-1089` | `df3e-6ef3`<br>`0c3f-1280`<br>`4756-47e0` | Energy storage / consumption tracking over 100s reflecting updated radio transceiver state transitions. |
     | **17** | `/showcases/wireless/ratecontrol/`<br>`-c NoRateControl -r 0` | `showcases.csv` | 14s | `tplx`<br>`~tNl`<br>`~tND`<br>`tyf` | `7ee9-503a`<br>`0816-e58f`<br>`648e-6e84`<br>`dad3-7f89` | `a9c1-6412`<br>`3302-4675`<br>`ce40-76a7`<br>`dad3-7f89` | Fixed-rate 802.11 transmission sequence with explicit frame mode tag dispatch (`findTag<Ieee80211ModeReq>()`). |
     | **18** | `/showcases/wireless/ratecontrol/`<br>`-c AarfRateControl -r 0` | `showcases.csv` | 12s | `tplx`<br>`~tNl`<br>`~tND`<br>`tyf` | `a7bc-05bb`<br>`9de0-4dd3`<br>`1209-101b`<br>`7539-d32d` | `00bd-a83c`<br>`be0d-7457`<br>`74b7-a035`<br>`7539-d32d` | AARF adaptive rate control responding to matched-filter SNIR noise scaling (`ScalarSnir`) and analog model band containment. |
     | **19** | `/tutorials/configurator/`<br>`-c Step10C -r 0` | `tutorials.csv` | 100s | `tplx`<br>`~tNl`<br>`~tND`<br>`tyf` | `71f5-b341`<br>`4a61-20ba`<br>`c614-3445`<br>`f078-56fe` | `360e-3e96`<br>`e262-7eb4`<br>`f0d2-98ee`<br>`f078-56fe` | Mixed wired/wireless tutorial scenario with auto-configured routes and wireless hosts under updated MAC/PHY event timing. |
     | **20** | `/tutorials/configurator/`<br>`-c Step12 -r 0` | `tutorials.csv` | 100s | `tplx`<br>`~tNl`<br>`~tND`<br>`tyf` | `e6ab-f59b`<br>`b3d9-9660`<br>`312a-1f31`<br>`4b03-e5a1` | `2c54-2317`<br>`203f-8319`<br>`3e16-4bf1`<br>`4b03-e5a1` | Multi-interface node configuration with 802.11 wlan and Ethernet interfaces undergoing updated MAC/PHY scheduling. |
     | **21** | `/examples/seaport/`<br>`-c General -r 0` | `examples.csv` | 1000s | `tplx`<br>`~tNl`<br>`tyf` | `32cb-156a`<br>`fcf4-00cd`<br>`e3e9-834d` | `d2b0-5f71`<br>`0271-eeef`<br>`e3e9-834d` | 8 mobile vessels communicating with 6 stationary APs over 1000s under `ScalarSnir` matched-filter noise scaling and receiver band containment. |

   - **Important Notes on Ingredient Composition Changes**:
     1. **Row 3 (`datalinkactivity Dynamic`) — Ingredient Addition**: The upstream baseline covered 3 ingredients (`tplx`, `~tNl`, `~tND`). The updated baseline adds the canvas visualizer graphical ingredient (`tyf` = `e720-1bfa`), expanding fingerprint verification to visual figure elements.
     2. **Row 6 (`routingtable Dynamic`) — Coverage Reduction**: The upstream baseline covered 4 ingredients (`tplx`, `~tNl`, `~tND`, `tyf`), whereas the updated baseline retains only `tplx` (`ef00-2b2b`). Ingredients `~tNl` and `~tND` invoke packet serialization checks on IPv4 headers; in this dynamic routing scenario without explicit checksum computation enabled, serialization fails with `cRuntimeError`. Rather than updating hashes, coverage was narrowed to `tplx`. This represents a structural coverage reduction that should be flagged for test suite maintenance.

   - **Re-Verification of Re-Baselined Suite**:
     - Command: `./fingerprinttest -d -m '(routing/manet|visualizer/canvas/(datalinkactivity|instrumentfigures|networkpathactivity|routingtable|statistic)|wireless/(analogmodel|blockack|fragmentation|power|ratecontrol)|tutorials/configurator.*Step1(0C|2)|seaport)' -f 'tplx' -f '~tNl' -f '~tND'`
     - Result: **100% PASS** (all 21 updated configurations match expected fingerprints).

3. **Simulation Examples**:
   - Verified configurations: `Ht20MHz`, `Ht40MHzSecondaryAbove`, and `Ht40MHzSecondaryAboveWithInterferer`.
   - Behavior: Stations accurately detect secondary channel interference, defer HT40 access per IEEE 802.11-2024 Clause 11.15.9, and achieve expected throughput.

4. **Architectural & Standards Compliance**:
   - Normative text verified against IEEE Std 802.11-2024 Clauses 11.15.9, 10.23.2.4, 19.3.19.6.1, 19.3.19.6.4, and 19.3.19.6.5.

PR #1144, opened 2026-08-29
IEEE 802.11: add HT capability and operation signalling

mgonzalezlopezudc:
## Summary

- Model and serialize HT Capabilities and HT Operation management elements.
- Derive legacy operational rates and HT capabilities from authoritative PHY mode sets.
- Negotiate and store per-peer HT state during association and discovery.
- Select peer-compatible HT modes using negotiated MCS, bandwidth, and short-GI support.
- Make AP response transactions, AID ownership, and STA lifecycle cleanup reliable.
- Correct PHY response timing and management-frame encoding details.

## Validation

- Debug build: `make MODE=debug -j$(nproc)`
- Unit tests: 13/13 passed
- IEEE 802.11 module tests: 13/13 passed
- Queueing tests: 1/1 passed
- Total focused tests: 27/27 passed

Fingerprint checks and updates were intentionally deferred to the last commit of the PR. After making that commit, Devin signaled a new issue, and we took care of it. That new issue had nothing to do with the fingerprint update: it was already present but for some reason Devin did not catch it before. 

---

<a href="https://app.devin.ai/review/inet-framework/inet/pull/1144" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=3">
    Devin Review
  </picture>
</a>


src/inet/linklayer/ieee80211/mgmt/Ieee80211MgmtFrameSerializer.h
```diff
@@ -17,20 +19,31 @@ namespace ieee80211 {
...
     virtual void serialize(MemoryOutputStream& stream, const Ptr<const Chunk>& chunk) const override;
-    virtual const Ptr<Chunk> deserialize(MemoryInputStream& stream) const override;
+    static const Ptr<Chunk> deserializeFrame(MemoryInputStream& stream, const std::type_info& typeInfo);
 
   public:
     Ieee80211MgmtFrameSerializer() : FieldsChunkSerializer() {}
 };
 
+template<typename Frame>
+class Ieee80211TypedMgmtFrameSerializer : public Ieee80211MgmtFrameSerializer
```
levy, line 36: why do we need to add template parameters to serializers? I think we should not, INET did work without template parameterized header serializers in the past. There should be a very good reason to introduce this.

src/inet/physicallayer/wireless/ieee80211/mode/Ieee80211ModeSet.cc
```diff
@@ -26,56 +27,56 @@ Register_Abstract_Class(Ieee80211ModeSet);
...
     Ieee80211ModeSet("a", {
-        { true, &Ieee80211OfdmCompliantModes::ofdmMode6MbpsCS20MHz },
-        { false, &Ieee80211OfdmCompliantModes::ofdmMode9MbpsCS20MHz },
-        { true, &Ieee80211OfdmCompliantModes::ofdmMode12MbpsCS20MHz },
-        { false, &Ieee80211OfdmCompliantModes::ofdmMode18MbpsCS20MHz },
-        { true, &Ieee80211OfdmCompliantModes::ofdmMode24MbpsCS20MHz },
-        { false, &Ieee80211OfdmCompliantModes::ofdmMode36Mbps },
-        { false, &Ieee80211OfdmCompliantModes::ofdmMode48Mbps },
-        { false, &Ieee80211OfdmCompliantModes::ofdmMode54Mbps },
+        { true, &Ieee80211OfdmCompliantModes::ofdmMode6MbpsCS20MHz, true },
```
levy, line 30: What is the last boolean doing here? It's always true and I didn't find it's purpose in the code. Maybe it is justified and I just need some clarification.

levy (commented):
Let's see how the reviewing process works.

src/inet/physicallayer/wireless/ieee80211/mode/Ieee80211ModeSet.h
```diff
@@ -16,6 +16,15 @@ namespace physicallayer {
 
 class INET_API Ieee80211ModeSet : public IPrintableObject, public cObject
 {
+  public:
+    enum class OperatingPhy {
```
levy: This name looks weird to me. Looking at the actual enum values suggests this should be named differently. What does operating mean in this enum? It would be the role where the enum is used and not the name of the class. Also, isn't this some PHY mode instead of just PHY, because an operating PHY seems like a PHY module.

src/inet/physicallayer/wireless/ieee80211/mode/Ieee80211ModeSet.cc
mgonzalezlopezudc: That last value is the `isLegacyOperational` flag in the `Entry` struct, not a duplicate of `isMandatory`.

You can see it in `Ieee80211ModeSet.h:29-37`:

- `bool isMandatory;`
- `const IIeee80211Mode *mode;`
- `bool isLegacyOperational = false;`

So the pattern:

- `{ true, &SomeMode, true }`

means:

- mandatory
- this mode object
- and this mode is eligible as a legacy operational / Supported Rates mode

The constructor then uses that flag to build the legacy operational list and validate it, as shown in `Ieee80211ModeSet.cc:494-555`:

- it collects entries where `entry.isLegacyOperational` is true
- it enforces that they have representable legacy rates
- it uses them for Supported Rates / operational-rate advertising

### Why it is “always true” there
Because those entries are intentionally modeled as legacy rates that must be advertised as operational/Supported Rates, not because the flag is redundant in general. In the same file, many HT/VHT MCS entries use `false` for that third field, which is exactly how the code distinguishes “supported mode” from “legacy operational mode”. See the HT/VHT entries in `Ieee80211ModeSet.cc:81-157`.

> In short: the last boolean means “this mode is part of the legacy operational rate set,” and it happens to be true for the legacy modes listed in those tables.

levy (commented):
another one

src/inet/queueing/base/PacketQueueBase.h
```diff
@@ -40,7 +41,15 @@ class INET_API PacketQueueBase : public PacketProcessorBase, public virtual IPac
 
     virtual void enqueuePacket(Packet *packet) override;
     virtual Packet *dequeuePacket() override;
+    virtual void setPacketDropCallback(IPacketQueue::ICallback *callback) override { packetDropCallback = callback; }
```
levy: INET already has a mechanism for signaling packet drops. When a packet is dropped, the module emits a packetDroppedSignal, see dropPacket call in PacketQueue, and also PacketProcessorBase::dropPacket. This new mechanism seems to be redundant for me and re-inventing something that already exists. Why don't we use the existing mechanims? Statistics and other model behavior already depends on packets being dropped, so it should be signalled by the model anyway.

src/inet/physicallayer/wireless/ieee80211/mode/Ieee80211ModeSet.h
mgonzalezlopezudc: _Codex says this and sounds reasonable to me, tell me if you agree:_

Yes—`OperatingPhy` is misleading. I’d rename the enum to `PhyType`, not `PhyMode`.

Here, “operating” only describes the value’s role: it is the PHY family associated with the selected mode set and is currently used to choose PHY-specific TXOP defaults. It does not represent an operating PHY module.

IEEE 802.11-2024 calls OFDM, HR/DSSS, ERP, HT, and VHT PHYs or PHY types. Table 9-194 groups TXOP limits by the clauses defining those PHYs, while Table 9-438 explicitly uses “PHY Type.”

`PhyMode` would be less accurate in INET because `IIeee80211Mode` already means a concrete transmission mode—bitrate/MCS, bandwidth, preamble, guard interval, and so on. For example, an HT mode set can contain both legacy and HT modes while its overall PHY type remains HT.

I suggest:

```cpp
enum class PhyType {
    OFDM,
    HR_DSSS,
    ERP,
    HT,
    VHT,
};

const PhyType operatingPhyType;
PhyType getOperatingPhyType() const;
```

That separates the domain name (`PhyType`) from the contextual role (`operatingPhyType`). The current declaration is at [Ieee80211ModeSet.h](/home/user/omnetpp_ws/inet-addba-transaction/src/inet/physicallayer/wireless/ieee80211/mode/Ieee80211ModeSet.h:23), and its only behavioral consumer is the TXOP selection in [TxopProcedure.cc](/home/user/omnetpp_ws/inet-addba-transaction/src/inet/linklayer/ieee80211/mac/originator/TxopProcedure.cc:34).

levy (commented):
I think the new packet drop signaling mechanism is redundant with the existing one. See PacketDropReason and it's usage in Simsignals.msg.

It feels like this patch re-invents something that's already there: modules signal dropping a packet along with its reason: queue is full, retry limit reached, etc

if we need a new reason we should simply add it.

src/inet/linklayer/ieee80211/mac/contract/IFrameTransmissionCallback.h
levy, line 1: I think this file should not exist, because it's whole functionality is covered by sending signals about dropped packets.

src/inet/linklayer/ieee80211/mac/framesequence/FrameSequenceContext.cc
```diff
@@ -78,4 +78,3 @@ void FrameSequenceNumPacketsFilter::receiveSignal(cResultFilter *prev, simtime_t
 
 } // namespace ieee80211
 } // namespace inet
-
```
levy, line 81: we should avoid unnecessary whitespace changes, because it shows up in the changed file list.

if we want to fix whitespaces, then it has to be a separate patch.

BTW, in general it's always a good idea to keep large local or global cross cutting whitespace changes as a separate commit.

src/inet/queueing/contract/IPacketBuffer.h
```diff
@@ -27,6 +27,13 @@ class INET_API IPacketBuffer : public virtual IPacketCollection
          * The packet is never nullptr.
          */
         virtual void handlePacketRemoved(Packet *packet) = 0;
+
```
levy: This should not be here for the same packet dropped signal reason.

src/inet/queueing/contract/IPacketQueue.h
```diff
@@ -20,6 +20,17 @@ namespace queueing {
  */
 class INET_API IPacketQueue : public virtual IPacketCollection, public virtual IPassivePacketSink, public virtual IPassivePacketSource
 {
+  public:
```
levy: This should not be here for the same packet dropped signal reason.

levy (commented):
New comments related to packet drop signalling.

src/inet/queueing/base/PacketQueueBase.h
mgonzalezlopezudc: I don't remember if this is the case, but usually this kind of "redundant" new mechanisms are introduced by the AI agent not on its first implementation but later on to solve bugs/flags signaled by Devin. I'm reviewing if we can avoid introducing this new redundant packet drop signaling mechanism

mgonzalezlopezudc: _Codex response (LGTM, tell me what you think):_

Verdict: accept [[Levy’s comment](https://github.com/inet-framework/inet/pull/1144#pullrequestreview-5064553763)](https://github.com/inet-framework/inet/pull/1144#pullrequestreview-5064553763). It is correct in substance.

- `PacketDropReason` already defines `QUEUE_OVERFLOW` and `RETRY_LIMIT_REACHED` in [Simsignals.msg](/home/user/omnetpp_ws/inet-feat-ht-cap-op-sig/src/inet/common/Simsignals.msg:11).
- `PacketProcessorBase::dropPacket()` emits `packetDroppedSignal` before deleting the packet, so listeners can synchronously inspect the live packet and transaction tag. The callback’s lifetime rationale therefore does not distinguish it from the signal: [PacketProcessorBase.cc](/home/user/omnetpp_ws/inet-feat-ht-cap-op-sig/src/inet/queueing/base/PacketProcessorBase.cc:387).
- The branch currently invokes the new callback and then immediately emits the existing signal through `dropPacket()`: [PacketQueue.cc](/home/user/omnetpp_ws/inet-feat-ht-cap-op-sig/src/inet/queueing/queue/PacketQueue.cc:102). This is duplicate notification.
- The callback carries less information, supports only one consumer, and adds a pure virtual method to the generic `IPacketQueue` contract.
- OMNeT++ signals propagate to listeners on ancestor components. `CompoundPacketQueueBase` already relies on this behavior to observe drops from child queues: [CompoundPacketQueueBase.cc](/home/user/omnetpp_ws/inet-feat-ht-cap-op-sig/src/inet/queueing/queue/CompoundPacketQueueBase.cc:17).

Recommended resolution:

1. Remove `IPacketQueue::ICallback`, `setPacketDropCallback()`, `notifyPacketDropped()`, and their queue/buffer propagation machinery.
2. Have DCF/HCF subscribe to `packetDroppedSignal` on their pending-queue components and handle tagged management frames in `receiveSignal()`.
3. Keep `IFrameTransmissionCallback`: it represents the higher-level terminal management outcome, including successful ACK completion, so it is not equivalent to a drop signal.
4. Do not add another `PacketDropReason`. Queue overflow already describes the cause; `DROPPED_BEFORE_TRANSMISSION` is the management transaction’s derived outcome.

The only caveat is an externally configured `PacketBuffer` outside the queue subtree: its drop signal originates from the buffer. Current 802.11 pending queues do not use such buffers, so the branch’s generic external-buffer callback and test are unnecessary scope for this fix. Supporting that configuration should be a separate signal-routing change.

This review does not require a change in `Ieee80211MgmtSta.cc`; the affected path is the AP response transaction through DCF/HCF pending queues.

src/inet/linklayer/ieee80211/mac/contract/IFrameTransmissionCallback.h
mgonzalezlopezudc: check https://github.com/inet-framework/inet/pull/1144#discussion_r3893106887

src/inet/linklayer/ieee80211/mgmt/Ieee80211MgmtFrameSerializer.h
mgonzalezlopezudc: _Codex response follows. In summary: eliminate through a cleanup. Tell me what you think about it._

The template is useful here, but only as an adapter around a limitation in `FieldsChunkSerializer`.

`Chunk::deserialize()` already knows and passes the requested concrete type ([Chunk.cc](https://github.com/mgonzalezlopezudc/inet/blob/cleanup/ht-cap-op-signalling/src/inet/common/packet/chunk/Chunk.cc#L200)). However, `FieldsChunkSerializer` discards that `typeInfo` and invokes a protected overload without it ([FieldsChunkSerializer.cc](https://github.com/mgonzalezlopezudc/inet/blob/cleanup/ht-cap-op-signalling/src/inet/common/packet/serializer/FieldsChunkSerializer.cc#L28)).

Consequently, the management-frame body cannot tell whether it should produce an Association Request, Beacon, Probe Response, etc. The template binds that missing information into each registered serializer:

```cpp
Ieee80211TypedMgmtFrameSerializer<Ieee80211BeaconFrame>
```

Its advantages are:

- preserves the requested concrete frame subtype;
- shares all actual serialization/deserialization code;
- requires no mutable type field or constructor argument;
- adds only a tiny compile-time adapter per registered subtype.

It is not providing generic serialization logic—the implementation still dispatches through `type_info` in [Ieee80211MgmtFrameSerializer.cc](https://github.com/mgonzalezlopezudc/inet/blob/cleanup/ht-cap-op-signalling/src/inet/linklayer/ieee80211/mgmt/Ieee80211MgmtFrameSerializer.cc#L566). It also is not completely type-safe: someone could accidentally register `FrameA` with `Ieee80211TypedMgmtFrameSerializer<FrameB>`.

My recommendation:

- For this focused branch, keep the template. The local non-template alternatives require duplicated subclasses, stored runtime type state, or duplicated `FieldsChunkSerializer` bookkeeping, all of which are worse.
- Long-term, eliminate it through a framework-level cleanup: change `FieldsChunkSerializer` so its protected deserialization hook receives the `typeInfo` already present in the public `ChunkSerializer` contract ([ChunkSerializer.h](https://github.com/mgonzalezlopezudc/inet/blob/cleanup/ht-cap-op-signalling/src/inet/common/packet/serializer/ChunkSerializer.h#L31)). Then all management types could register the same non-template `Ieee80211MgmtFrameSerializer`.

So: avoiding the template would produce a cleaner final API, but only if the underlying `FieldsChunkSerializer` interface is corrected. Avoiding it solely inside this serializer would not be an improvement.

src/inet/physicallayer/wireless/ieee80211/mode/Ieee80211ModeSet.h
levy: I agree with your suggestion.

mgonzalezlopezudc:
The refactor commit https://github.com/inet-framework/inet/pull/1144/commits/ffe40eda3501ea77814aecc4cc92cfb30b5c3b12 made Devin to flag another bug (that was already present before the refactor commit). It's been corrected in commit https://github.com/inet-framework/inet/pull/1144/commits/3be545a22ffe23218edb0f34fb2060f14d8ea4a3

PR #1154, opened 2026-09-01
IEEE 802.11: add HT capability and operation signalling

mgonzalezlopezudc:
## Dependency

Depends on #1166 (`ethernet: preserve fragment FCS type when deserializing`) at `353798aea4719c2823cc9da474e27c48eaabc87b`.

PR #1166 is itself stacked on #1164. Until #1166 is merged, GitHub includes its 10 prerequisite commits in this PR’s comparison against `master`. The HT topic itself is the 29-commit range:

```text
353798aea4719c2823cc9da474e27c48eaabc87b..bbc7e20ed7698bde82c603d1e4f49f340a43d9a7
```

## Summary

This PR adds model-backed IEEE 802.11 HT Capabilities and HT Operation signalling and uses the negotiated state throughout discovery, association, channel validation, and rate selection.

It also:

- completes the mixed HT mode catalogue and makes PHY timing references explicit;
- encodes legacy operational rates and management fixed fields correctly;
- derives advertised HT widths from the operational transmitter/receiver intersection;
- carries HT state through Beacon, Probe, Association, and Reassociation exchanges;
- restricts unicast rate selection to modes supported by negotiated peer state;
- makes STA association, reassociation, teardown, and restart transitions explicit;
- commits AP association state and AIDs only after the exact response transaction succeeds;
- handles queue drops, retry exhaustion, RTS failure, fragmentation, and HCF internal collisions as terminal transaction outcomes;
- exposes terminal frame-transmission outcomes through a signal and result recorders;
- refreshes associated-peer HT state from accepted Beacons;
- revalidates HT operation and negotiated peer state after runtime channel or band changes;
- rejects malformed or contradictory peer HT advertisements without terminating the simulation.

This is the review-ready reconstruction of the work from #1144 as a linear, dependency-ordered 29-commit topic series.

## Motivation

The previous model did not represent and negotiate HT capability and operation state end to end. In particular:

- operational rate membership was inferred incompletely;
- HT mode timing and channel-width authority were ambiguous;
- internal radio channel indices could be confused with IEEE on-wire channel numbers;
- unicast rate selection could choose an HT mode unsupported by the peer;
- association state and AIDs could become effective before the corresponding response was acknowledged;
- terminal queue and MAC failure paths could leave association transactions pending;
- runtime channel changes could leave advertised and negotiated HT state stale;
- malformed peer elements could abort the simulation or partially update cached state;
- one-antenna radios could advertise unsupported multi-stream MCS values.

The implementation makes these authorities and transitions explicit while preserving legacy operation for peers without negotiated HT support.

## Reading order

The 29 topic commits are ordered by dependency:

1. **PHY and wire-format foundations (C01–C05):** timing authorities, operational rates, typed HT state, the complete mixed-mode catalogue, and operational PHY widths.
2. **Negotiation and STA behavior (C06–C14):** local transaction metadata, detailed association transitions, HT discovery and association negotiation, peer-compatible rate selection, fixed-field byte order, channel validation, and lifecycle cleanup.
3. **Reliable AP completion (C15–C18):** frame-transmission completion, compound-queue ownership, transactional AP responses, and EDCAF retry ownership.
4. **Robustness and observability (C19–C24):** malformed peer handling, reassociation completion, antenna-stream limits, transmission-result signalling, Beacon-driven peer-state refresh, and AP restart cleanup.
5. **Runtime consistency and validation (C25–C29):** resolution-independent timestamp coverage, dynamic channel and band revalidation, pending-association reconciliation, AP-side HT validation, and contradictory transmit-field rejection.

Tests and fingerprint updates travel with their causal commits.

### Contracts

- Adds packet-PHY channel and HT-width provider contracts.
- Publishes terminal MAC frame-transmission outcomes through `frameTransmissionFinishedSignal`.
- Carries immutable `FrameTransmissionDetails` and `FrameTransmissionStatus` values to management and non-behavioral observers.
- Keeps management transaction identity sender-local and removes it at the wireless packet-domain boundary.
- Uses radio channel-change notifications to keep AP HT operation and peer state synchronized.

### Packet representation

- Adds typed HT Capabilities and HT Operation management elements.
- Preserves Supported and Extended Supported Rate membership.
- Encodes management fixed-width fields little-endian.
- Validates Association ID marker bits and range.
- Converts internal channel indices to and from IEEE channel numbers at the management boundary.
- Safely rejects invalid HT channel tuples, unsupported Basic MCS sets, and contradictory transmit-MCS fields.
- Extends coverage for the microsecond Beacon and Probe Response timestamp encoding supplied by the prerequisite stack.

### State and ownership

- The MIB owns local and per-peer HT capability, operation, association-ID, and station-membership state.
- Detailed STA management owns pending association and reassociation transitions.
- AP association state becomes effective only after the correlated response is acknowledged.
- Runtime channel changes revalidate current and pending peer HT state.
- Accepted Beacons refresh the authoritative associated-AP snapshot; invalid updates preserve the existing association and fall back safely.
- Compound queues retain packet ownership while synchronously reporting drops.

### Configuration and observability

- Adds MIB parameters for the modeled HT A-MPDU exponent, secondary-channel offset, and protection mode.
- Adds explicit MIB and radio module references where required.
- Adds `htNegotiationFailed` diagnostics.
- Adds `frameAcked`, `frameRetryLimitReached`, and `frameDroppedBeforeTx` statistics.

No changed topic source path is sealed. No new architecture or naming exception is required.

## Validation

The topic was replayed and verified across eight upstream checkpoints while reusing one debug object tree. The final safe point is `bbc7e20ed7698bde82c603d1e4f49f340a43d9a7`.

Build commands:

```bash
make MODE=debug -j$(nproc) LN='cp -f'
make -C tests/unit/lib MODE=debug -j$(nproc)
make -C tests/module/lib MODE=debug -j$(nproc)
```

Result: **PASS**. The final incremental build completed without cleaning the object tree.

Focused unit tests:

```bash
UNIT_FILTER='Ieee80211HtModeSet_1|Ieee80211ResponseTimeout_1|Ieee80211TxopProcedure_1|Ieee80211SupportedRates_1|Ieee80211HtCapabilities_1|Ieee80211HtMgmtElements_1|Ieee80211MgmtProtocolPrinter_1|Ieee80211MgmtTransactionTag_1|Ieee80211MgmtStaPrimitiveDispatch_1|Ieee80211MgmtStaDiscovery_1|Ieee80211PeerModeSelection_1|Ieee80211MgmtFrameSerializer_1|Ieee80211Band_1|Ieee80211MgmtApTransaction_1|Ieee80211MibAssociationId_1|FieldsChunkSerializer_1|EthernetMacHeaderSerializer_1'

printf "run_opp_tests(test_folder='tests/unit', filter='$UNIT_FILTER', mode='debug', build=False)\nexit\n" |
  env INET_ROOT="$PWD" PATH="$PWD/bin:$PATH" opp_repl --load @opp -p inet
```

Result: **17/17 PASS**.

Focused module tests:

```bash
MODULE_FILTER='Ieee80211PacketDomainTagBoundary_1|Ieee80211MgmtStaDiscovery_1|Ieee80211MgmtStaDeauthentication_1|Ieee80211MgmtStaDisassociation_1|Ieee80211MgmtStaLifecycle_1|Ieee80211HtAssociation_1|Ieee80211MgmtApHcfQueueDrop_1|Ieee80211MgmtApHcfRtsTimeout_1|Ieee80211MgmtApQueueDrop_1|Ieee80211MgmtApReassociationSnapshot_1|Ieee80211MgmtApTimeout_1|Ieee80211MgmtApUnavailableChannel_1|Ieee80211MgmtStaSimplifiedInitialization_1|Ieee80211HcfInternalCollision_1|Ieee80211AgentStaReassociation_1|Ieee80211HtAntennaRateControl_1|Ieee80211FrameTransmissionStats_1|Ieee80211MgmtStaBeaconUpdate_1|Ieee80211MgmtApLifecycle_1|Ieee80211MgmtApChannelChange_1|Ieee80211MgmtApMalformedHtCap_1|EthernetFragmentFcsChecker_1'

printf "run_opp_tests(test_folder='tests/module', filter='$MODULE_FILTER', mode='debug', build=False)\nexit\n" |
  env INET_ROOT="$PWD" PATH="$PWD/bin:$PATH" opp_repl --load @opp -p inet
```

Result: **22/22 PASS**.

Focused queueing test:

```bash
printf "run_opp_tests(test_folder='tests/queueing', filter='CompoundQueue_1', mode='debug', build=False)\nexit\n" |
  env INET_ROOT="$PWD" PATH="$PWD/bin:$PATH" opp_repl --load @opp -p inet
```

Result: **1/1 PASS**.

The following reproduces the exact 37 fingerprint rows changed by this topic relative to PR #1166:

```bash
fingerprint_rows=$(mktemp)

git diff --unified=0 \
  353798aea4719c2823cc9da474e27c48eaabc87b..HEAD -- \
  tests/fingerprint/examples.csv \
  tests/fingerprint/showcases.csv \
  tests/fingerprint/tutorials.csv |
  sed -n '/^+++ /d; /^+/s/^+//p' > "$fingerprint_rows"

test "$(wc -l < "$fingerprint_rows")" -eq 37

(
  cd tests/fingerprint
  ./fingerprinttest \
    -d \
    -q \
    -m '.*' \
    -f tplx \
    -f '~tNl' \
    -f '~tND' \
    "$fingerprint_rows"
)

rm "$fingerprint_rows"
```

Result: **37/37 PASS**.

Source and history checks:

```bash
git diff --check 353798aea4719c2823cc9da474e27c48eaabc87b..HEAD
doc/project/enforcement/check-source-seals.sh \
  --base 353798aea4719c2823cc9da474e27c48eaabc87b \
  --head HEAD
doc/project/enforcement/check-architecture.sh src/inet/linklayer/ieee80211
doc/project/enforcement/check-architecture.sh src/inet/physicallayer/wireless/ieee80211
doc/project/enforcement/check-commits.sh \
  353798aea4719c2823cc9da474e27c48eaabc87b..HEAD
```

The branch is linear with exactly 29 topic commits and no merge commits. Seal, architecture, whitespace, baseline-coupling, ordering, and rename checks pass.

The commit-message check has one known metadata finding: C26 (`d0548c4433`) has a 73-character subject, one character over the approximate 72-character limit.

## Fingerprint updates

This topic updates 37 existing fingerprint rows relative to PR #1166:

- 17 in `tests/fingerprint/examples.csv`;
- 16 in `tests/fingerprint/showcases.csv`;
- 4 in `tests/fingerprint/tutorials.csv`.

The updates remain in their causal commits:

- C01 moves seven rows due to the corrected PHY timing authority;
- C02 moves the remaining 30 rows due to operational-rate encoding;
- C10 advances the TXOP row after negotiated peer-rate selection;
- C11 advances the same 30 C02 rows to their final management-wire-byte fingerprints.

The staged rebase reproduced the expected timestamp transition exactly before the prerequisite baseline checkpoint and passed the combined oracle afterward. No additional repository baseline was introduced by the rebase.

## Scope limits

The current packet PHY advertises operational HT20 only. HT40 catalogue, negotiation, selection, and channel validation are covered, but an end-to-end HT40 PHY exchange is outside this PR.

VHT capability negotiation is also outside the scope of this change.

---

<a href="https://app.devin.ai/review/inet-framework/inet/pull/1154" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=3">
    Devin Review
  </picture>
</a>


src/inet/linklayer/ieee80211/mgmt/Ieee80211MgmtFrameSerializer.cc
```diff
@@ -7,23 +7,167 @@
...
-Register_Serializer(Ieee80211AssociationResponseFrame, Ieee80211MgmtFrameSerializer);
-Register_Serializer(Ieee80211AuthenticationFrame, Ieee80211MgmtFrameSerializer);
-Register_Serializer(Ieee80211BeaconFrame, Ieee80211MgmtFrameSerializer);
-Register_Serializer(Ieee80211DeauthenticationFrame, Ieee80211MgmtFrameSerializer);
-Register_Serializer(Ieee80211DisassociationFrame, Ieee80211MgmtFrameSerializer);
-Register_Serializer(Ieee80211ProbeRequestFrame, Ieee80211MgmtFrameSerializer);
-Register_Serializer(Ieee80211ProbeResponseFrame, Ieee80211MgmtFrameSerializer);
-Register_Serializer(Ieee80211ReassociationRequestFrame, Ieee80211MgmtFrameSerializer);
-Register_Serializer(Ieee80211ReassociationResponseFrame, Ieee80211MgmtFrameSerializer);
+Register_Serializer(Ieee80211AssociationRequestFrame, Ieee80211TypedMgmtFrameSerializer<Ieee80211AssociationRequestFrame>);
```
levy: I find using C++ templates for header serializers unnecessary. What's the reason behind them?
We could live without having C++ templates for a long time, even including message type aware deserialization into the specific C++ type. I have a strong feeling that serialization should not be done this way.

So what's the issue it's solving?

src/inet/physicallayer/wireless/ieee80211/bitlevel/Ieee80211LayeredOfdmReceiver.cc
```diff
@@ -411,6 +412,14 @@ const IReceptionResult *Ieee80211LayeredOfdmReceiver::computeReceptionResult(con
     }
     // add indications
     auto packet = const_cast<Packet *>(packetModel->getPacket());
+    // Packet-domain error models duplicate the transmitted packet, including
```
levy, line 415: Hmm, I'm not saying this wrong or anything, but the Ieee80211LayeredOfdmReceiver class is only used in the bit level simulation model and it's not used at all in the packet level model.

So why is there no change related to this in the packet level simulation model? Is that already correct or not affected?

src/inet/linklayer/ieee80211/mac/contract/IFrameTransmissionCallback.h
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Reports terminal outcomes for locally originated frame transmissions.
+ *
+ * The result is created synchronously by the MAC and is valid only for the
+ * duration of the callback. In particular, the frame pointer must not be
+ * retained by the receiver: the coordination function may release the frame
+ * immediately after the callback returns.
+ */
+class INET_API IFrameTransmissionCallback
```
levy: This callback seems like a signal that should be emitted. The main question is this: are we sure there will ever be only one interesting party? It's perfectly fine using signals between internal parts of IEEE 802.11 components for communication.

I'm not sure about this signal thing though, I'm open for discussion.

levy (commented):
Please check these and let's discuss!

src/inet/physicallayer/wireless/ieee80211/bitlevel/Ieee80211LayeredOfdmReceiver.cc
mgonzalezlopezudc: This clears all the tags attached to the transmitted packet to make sure the receiver only sees the actual transmitted packet, with the exception of the PacketProtocolTag. I think the receiver should not even see this one, but it's already common practice in INET as far as I remember.

src/inet/linklayer/ieee80211/mac/contract/IFrameTransmissionCallback.h
mgonzalezlopezudc: GPT-5.6 Sol says the following and it makes sense to me (i.e. keep it as is until a new observer of this event arises). My only doubt is if it can be interesting recording this event for statistics.  

---
Short answer: `IFrameTransmissionCallback` is not universally better than a signal. It is better for the current state-changing, single-recipient handoff; signals are better for observation and fan-out.

Current production topology confirms one behavior-driving recipient:

- The MAC stores one callback pointer, resolved from its sole `mgmtOut` peer: Ieee80211Mac.cc#L92.
- `Ieee80211MgmtAp` is the only production implementation: Ieee80211MgmtAp.h#L24.
- That receiver owns the association state and uses the result to commit/release the AID and clear the pending transaction: Ieee80211MgmtAp.cc#L98.

The callback gives this control path:

- A typed result with three exhaustive outcomes.
- An explicit, synchronous recipient.
- A documented borrowed-frame lifetime.
- No subscription-scope, source-filtering, or listener-order ambiguity.
- A complete result contract: `packetDroppedSignal` cannot report `ACKNOWLEDGED`.

This aligns with direct typed coordination in AR-COM-DIRECT and keeps association state under its single management owner.

**Are we sure only one party is interested?** We are sure there is only one current production party that changes protocol state. We cannot assume there will never be multiple observers. Statistics, tracing, visualization, or diagnostics could reasonably want the same event.

Signals are valid for those consumers. AR-COM-NOTIFY explicitly permits zero or more independent behavioral listeners for a completed fact. If such a consumer appears, I recommend emitting a declared terminal-transmission signal alongside the callback, exactly once and after the owner’s invariants are restored.

Therefore my recommendation is:

- Keep `IFrameTransmissionCallback` for the management control handoff.
- Do not add a signal speculatively.
- Add an observational signal later if an actual independent consumer needs it.
- Reconsider replacing the callback only if a second independent behavior-driving consumer appears.

src/inet/linklayer/ieee80211/mgmt/Ieee80211MgmtFrameSerializer.cc
mgonzalezlopezudc: I had already reviewed this issue before the cleanup. GPT-5.6 Sol argued the following. I might be missing something, but the alternative would be considering ten subclasses (i.e. inheritance Vs templating). Would that be preferable?

---
The templates are not required by IEEE 802.11 or INET architecture. I used them as a localized adapter for correct deserialization.

Previously, the serializer did not actually deserialize the requested subtype: it contained `switch (0) // TODO receive and dispatch on type_info parameter`, so every call followed the Association Request branch. That latent bug became relevant once the new tests exercised deserialization for multiple management-frame types.

The constraint is that:

- `ChunkSerializer::deserialize()` receives the requested `typeInfo`.
- `FieldsChunkSerializer`#L29 discards it and calls the parameterless protected `deserialize(stream)`.
- A management-frame body cannot infer its subtype from its own bytes; that subtype is encoded in the separate MAC header.
- Therefore `Ieee80211TypedMgmtFrameSerializer<Frame>` bakes the registered subtype into each serializer instance and passes `typeid(Frame)` to the shared implementation. The registrations are in `Ieee80211MgmtFrameSerializer.cc`#L23.

So “they weren’t needed before” is really “deserialization was previously incomplete.” Serialization itself still does not need the template because it can inspect the concrete chunk.

A cleaner long-term solution would be to make `FieldsChunkSerializer` expose a type-aware protected deserialization hook. Then one ordinary management serializer could receive `typeInfo` directly. That would change shared serializer infrastructure, whereas the template kept this fix local. The other local alternatives—ten handwritten subclasses or duplicating `FieldsChunkSerializer`’s bookkeeping—would be worse.

If you meant the separate `writeSupportedRateElements<Frame>` helper: that template merely shares code across sibling frame classes whose `supportedRates` field is not declared in their common base. It is convenient, but not fundamentally necessary.

levy: Hmm, good point.

I think this should be fixed in the packet API, because that's the correct thing to do.
Then the templates are not needed.
The packet API fix needs another PR, and I think we should be backward compatible there, because it's easy to do.

src/inet/physicallayer/wireless/ieee80211/bitlevel/Ieee80211LayeredOfdmReceiver.cc
levy: That is true, but it's suspicious to me that it's done here.

It's common practice on the sender side and this isn't that.

src/inet/linklayer/ieee80211/mac/contract/IFrameTransmissionCallback.h
levy: Actually if you think about this, these are absolutely the event kinds which can be useful to collect as statistics. I mean the 3 outcomes are important enough to at least be able to see them as a count or something, no?

levy (commented):
I answered to each 3.

src/inet/common/Simsignals.cc
```diff
@@ -22,6 +22,7 @@ simsignal_t l2ApDisassociatedSignal = cComponent::registerSignal("l2ApDisassocia
 simsignal_t linkBrokenSignal = cComponent::registerSignal("linkBroken");
 
 simsignal_t modesetChangedSignal = cComponent::registerSignal("modesetChanged");
+simsignal_t frameTransmissionFinishedSignal = cComponent::registerSignal("frameTransmissionFinished");
```
levy: We already have a transmissionEndedSignal, how is this different? I assume it is different but then it should have a more specific name because it's hard to distinguish.

src/inet/physicallayer/wireless/ieee80211/contract/packetlevel/IIeee80211HtChannelWidthProvider.h
```diff
@@ -0,0 +1,32 @@
...
+namespace inet {
+namespace physicallayer {
+
+using namespace inet::units::values;
+
+/**
+ * Reports HT channel widths that an IEEE 802.11 PHY component can actually
+ * operate, independently of the modes described by its mode catalog.
+ */
+class INET_API IIeee80211HtChannelWidthProvider
```
levy, line 21: I'm not sure about this interface. First, are we sure we want to have a separate interface for answering the channel width support question? What if there will be future questions for the receiver/transmitter related to HT? Are we going to have a separate interface for each one?

Maybe the name is not right, because it should be more generic like capabilities... I don't know, it just feels wrong. We should discuss.

src/inet/physicallayer/wireless/ieee80211/bitlevel/Ieee80211LayeredOfdmReceiver.cc
levy, line 415: But should packet-domain error models really produce a packet which carries the tags from original transmitted packet? Or should we rather strip the packets there? It feels a bit late here.

All error model callers should do the same?

src/inet/physicallayer/wireless/ieee80211/contract/packetlevel/IIeee80211ChannelProvider.h
```diff
@@ -0,0 +1,32 @@
...
+
+namespace inet {
+namespace physicallayer {
+
+class Ieee80211Channel;
+
+/**
+ * Provides the current IEEE 802.11 channel of a packet-level PHY component.
+ */
+class INET_API IIeee80211ChannelProvider
```
levy, line 21: Is this yet another interface that needs to be added due to the duality of receiver/transmitter?
Or is this just transmitter? Why don't we simply add the method there?

src/inet/common/Simsignals.cc
mgonzalezlopezudc: Done

mgonzalezlopezudc:
Already merged into master through other PRs

PR #1172, opened 2026-09-08
IEEE 802.11: negotiate and maintain STA HT associations

mgonzalezlopezudc:
Association responses must match the pending procedure and peer, and HT negotiation must reject malformed advertisements without corrupting discovery or retained associations. Add correlated STA transactions, validated discovery, lifecycle cleanup, committed-state reassociation handling, peer-aware rates, AID wire validation, and simplified association restoration.

## Stack and reading order

PR 2/4. Head: `pr/ht-02-sta-association`. Target: `master` (PR 1/4 [#1167](https://github.com/inet-framework/inet/pull/1167) merged). Range: `b61cd671d429..1ded7e8cdd1b`. Read the following commits in order:

- `441ff2701b` — ieee80211: keep management transaction metadata local
- `fe876073fc` — ieee80211: model detailed association transactions
- `2108cb5ec5` — ieee80211: negotiate HT from validated discovery state
- `80d1c66987` — ieee80211: preserve peer ownership during STA teardown
- `cfcd73de3b` — ieee80211: report reassociation completion from committed state
- `74b40b81b8` — ieee80211: select rates from negotiated HT capabilities
- `6627fcf3f1` — ieee80211: validate association ID wire encoding
- `1ded7e8cdd` — ieee80211: restore simplified associations across lifecycle changes

## Architectural surface

Local transaction tags and fragmentation/receive boundaries, STA management primitives and lifecycle, peer HT state, DCF/HCF rate selection, and Association ID serialization.

## Baselines

C16 (`74b40b81b8`) carries the TXOP fingerprint transition caused by peer-aware rate selection. C17 (`6627fcf3f1`) carries thirty data-byte transitions for corrected Association ID encoding. No other baseline changes are included.

## Verification

C11–C12 have passing per-commit debug builds and scoped opp_repl tests. C13–C18 and this PR tip were not tested separately, per the requested final-only verification. The complete stack passed the final build and test union.

The full-stack recorded build command was `make MODE=debug -j$(nproc)`. The scoped unit/module/queueing runs used `opp_repl --load @opp -p inet`, with exact `run_opp_tests` calls in the retained `.final.tests.py` execution artifact. Fingerprints used `fingerprinttest -d -q -f tplx -f "~tNl" -f "~tND"` with the retained explicit 37-row CSV. Exact commands, selectors, statuses, and source-tree mappings are in the accompanying execution report.

---

<a href="https://app.devin.ai/review/inet-framework/inet/pull/1172" target="_blank">
  <picture>
[...]

PR #1173, opened 2026-09-09
IEEE 802.11: complete AP association transactions from MAC outcomes

mgonzalezlopezudc:
# Complete AP association transactions from MAC outcomes

Enqueuing an association response does not establish successful delivery. Publish terminal MAC outcomes and use transaction identity to commit or release AP AIDs and negotiated state. Cover queue drops, retry exhaustion, malformed requests, configured MAC subscription, AP shutdown, and GenericRadio non-HT compatibility; charge internal collisions to the losing EDCAF. Synchronize accumulated baseline drift in the JSON fingerprint store with accepted CSV values.

## Stack and reading order

PR 3/4. Head: `pr/ht-03-ap-transactions`. Target: `master` (PR 1/4 [#1167](https://github.com/inet-framework/inet/pull/1167) and PR 2/4 [#1172](https://github.com/inet-framework/inet/pull/1172) merged). Range: `814e6a7438eb..095c457c10`. Read the following commits in order:

- `5c630b95f3` — ieee80211: preserve receiver header trailing spacing
- `b61cacb44c` — ieee80211: order MAC PHY includes by name
- `3c00448c1b` — ieee80211: explain incompatible HT transmitter and receiver types
- `e1bdc650d8` — ieee80211: use management headers for HCF retry accounting
- `42dd47f450` — ieee80211: publish terminal management frame outcomes
- `9f70afa7bd` — ieee80211: reserve association IDs before committing membership
- `726153fa00` — ieee80211: advertise AP HT discovery state from the MIB
- `c886bccbc0` — ieee80211: commit AP associations after response completion
- `9d0104c0a5` — ieee80211: charge collision retries to the owning EDCAF
- `095c457c10` — tests: synchronize stale JSON fingerprints with CSV baselines

## Architectural surface

Synchronous `frameTransmissionOutcome` signal (declared on DCF and HCF) and details (`FrameTransmissionDetails`), result filters/statistics, DCF/HCF completion reporting via upward signal propagation to MAC subscribers, AP pending transactions, MIB AID allocation, and GenericRadio non-HT fallback (retaining legacy unknown channel when no channel is published and reading HT operation only when HT is supported). No sealed packet-core paths changed.

## Baselines

C10 (`095c457c10`) synchronizes 72 obsolete run-0 tplx entries in `tests/fingerprint/store.json` with the accepted CSV baselines already present at `master` (`814e6a7438eb`). This addresses accumulated baseline-store synchronization drift from previous merges; no single source commit in this series caused the movement, and the source tree remains completely unmodified by this update.

Note on `check-commits.sh`: The mechanical checker flags C10 under `PR-SPLIT-BASELINE` solely because a baseline-only commit follows source commit C09 in the series; the commit body details the accumulated drift justification per `PR-SPLIT-BASELINE` and `TR-BASELINE-PROVENANCE`.

## Verification

Passing final debug build and passing filtered `opp_repl` test suite:
- `make MODE=debug -j$(nproc)`: PASS, exit 0.
- Unit and module tests: 2 unit tests (`Ieee80211MibAssociationId_1.test`, `Ieee80211MgmtApTransaction_1.test`) and 12 module tests (frame transmission statistics, AP queue drops, HCF queue drops, RTS timeouts, malformed HT capability rejection, timeouts, lifecycle stop/start, HT associations, antenna/rate control, internal collisions, GenericRadio AP discovery, and unavailable channel) PASS with exit 0.
- Architecture rules for `linklayer/ieee80211` and `physicallayer/wireless/ieee80211` PASS, exit 0.
- Fingerprints: All 72 synchronized fingerprint test cases pass with exit 0 under `opp_repl` in debug mode (run 0, tplx); aggregate selector of 311 tasks yielded 227 PASS, 0 FAIL, 0 ERROR, and 84 unrecorded baseline SKIPs.
- Working tree clean, exact tree match against preserved topic tree (`git diff --exit-code adf266f48f HEAD --`: exit 0; tree `00f9d50599`).

### DIFF
--- a/src/inet/linklayer/ieee80211/mac/Ieee80211Mac.cc
+++ b/src/inet/linklayer/ieee80211/mac/Ieee80211Mac.cc
@@ -74,8 +74,10 @@ void Ieee80211Mac::initialize(int stage)
                 modeSet->getMaximumNumberOfSpatialStreams());
         std::set<Hz> operationalChannelWidths;
         if (modeSet->isHtOperationSupported()) {
-            const auto *transmitter = check_and_cast<const Ieee80211Transmitter *>(radio->getTransmitter());
-            const auto *receiver = check_and_cast<const Ieee80211Receiver *>(radio->getReceiver());
+            const auto *transmitter = dynamic_cast<const Ieee80211Transmitter *>(radio->getTransmitter());
+            const auto *receiver = dynamic_cast<const Ieee80211Receiver *>(radio->getReceiver());
+            if (transmitter == nullptr || receiver == nullptr)
+                throw cRuntimeError("HT operation requires Ieee80211Transmitter and Ieee80211Receiver");
             for (auto channelWidth : modeSet->getHtSupportedChannelWidths())
                 if (transmitter->isHtChannelWidthSupported(channelWidth) &&
                         receiver->isHtChannelWidthSupported(channelWidth))
--- a/src/inet/linklayer/ieee80211/mac/Ieee80211Mac.cc
+++ b/src/inet/linklayer/ieee80211/mac/Ieee80211Mac.cc
@@ -23,7 +23,6 @@
 #include "inet/linklayer/ieee80211/mac/Ieee80211SubtypeTag_m.h"
 #include "inet/linklayer/ieee80211/mac/Rx.h"
 #include "inet/linklayer/ieee80211/mac/contract/IContention.h"
-#include "inet/linklayer/ieee80211/mac/contract/IFrameSequence.h"
 #include "inet/linklayer/ieee80211/mac/contract/IRx.h"
 #include "inet/linklayer/ieee80211/mac/contract/ITx.h"
 #include "inet/networklayer/contract/IInterfaceTable.h"
@@ -38,6 +37,8 @@ namespace ieee80211 {
 using namespace inet::physicallayer;
 
 Define_Module(Ieee80211Mac);
+
+simsignal_t Ieee80211Mac::frameTransmissionOutcomeSignal = cComponent::registerSignal("frameTransmissionOutcome");
 
 Ieee80211Mac::Ieee80211Mac()
 {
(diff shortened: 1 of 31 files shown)


src/inet/linklayer/ieee80211/mac/Ieee80211Mac.cc
```diff
@@ -392,6 +393,14 @@ void Ieee80211Mac::sendDownPendingRadioConfigMsg()
     }
 }
 
+void Ieee80211Mac::notifyFrameTransmission(const Packet *frame, FrameTransmissionStatus status)
```
levy: I'm not sure about this. Calling notifyFrameTransmission on Ieee80211Mac instead of emitting the frameTransmissionOutcomeSignal from the channel function module seems unnecessary. Signals can be emitted by the model where the thing happens, e.g. Dcf, Hcf. The subscriber can subscribe on any parent module the MAC or the interface, it will still get the signal notification. That saves us from having a public method which is called only for emitting a signal by other modules.

src/inet/linklayer/ieee80211/mgmt/Ieee80211MgmtAp.h
```diff
@@ -75,11 +91,20 @@ class INET_API Ieee80211MgmtAp : public Ieee80211MgmtApBase
     using Ieee80211MgmtApBase::receiveSignal;
     virtual void receiveSignal(cComponent *source, simsignal_t signalID, cObject *obj, cObject *details) override;
 
+    virtual void frameTransmissionFinished(const Packet *responseFrame, FrameTransmissionStatus status);
```
levy: Why is this a public method, when it's only called internally?
Now that I look around, I see quite a lot of public functions which may not need to be public at all.
As a general rule the visibility should be minimal.
Also, functions should be virtual if they are either overridden or expected to be overridden in the future, so they are a real extension point.

PR #1196, opened 2026-09-12
Fix SharingRegionTagSet::removeTagsWherePresent<T>() wiping tags of every type

### LINKED ISSUE #1184: removeTagsWherePresent<T>() removes region tags of every type, not just T
**Type:** bug
**Component:** `src/inet/common/packet/tag`
**Version:** INET 4.7.0 (`dfe270b21f`, 2026-07-07)

## The bug

`SharingRegionTagSet::removeTagsWherePresent<T>()` filters its return value by `T`, but not its removal:

```cpp
template<typename T>
inline std::vector<SharingRegionTagSet::RegionTag<T>> SharingRegionTagSet::removeTagsWherePresent(b offset, b length)
{
    auto result = getAllTags<T>(offset, length);
    clearTags(offset, length);          // <-- not templated
    return result;
}
```

`clearTags()` (`SharingRegionTagSet.cc`) truncates, splits or erases every region intersecting `[offset, offset+length)` with no type check anywhere in the loop. So the call removes **all** region tags over the range, returns only those of type `T`, and the caller never learns the rest are gone.

It reaches users through `Chunk::removeTagsWherePresent<T>()` (`Chunk.h:810`) and `Packet::removeRegionTagsWherePresent<T>()` (`Packet.h:1469`), which both forward straight to it.

The single-tag operations next to it are type-correct - `removeTag<T>()` and `removeTagIfPresent<T>()` go through `getTagIndex(typeid(T), ...)` - so the inconsistency looks accidental.

## Reproduction

```cpp
SharingRegionTagSet tags;
tags.addTag<CreationTimeTag>(b(0), b(1000))->setCreationTime(42);
tags.addTag<PropagationTimeTag>(b(0), b(1000));   // any second type

auto removed = tags.removeTagsWherePresent<CreationTimeTag>(b(0), b(1000));

ASSERT(removed.size() == 1);        // passes
ASSERT(tags.getNumTags() == 1);     // FAILS: 0 -- the PropagationTimeTag is gone too
```

## Affected call sites

**`src/inet/common/ResidenceTimeMeasurer.cc:55` and `:75`**

```cpp
packet->removeRegionTagsWherePresent<ResidenceTimeTag>(offset, length);
```

with `offset = b(0)`, `length = packet->getDataLength()`, so it clears the packet's whole region tag set over the data range. `PacketEventTag` (`FlowMeasurementStarter.cc:90`) and `FlowTag` (`FlowTag.cc:17`) live in that same set, so running residence-time measurement alongside flow measurement destroys the flow tags at every `measurementEnd` and every `packetDropped`.

**`src/inet/applications/tcpapp/TcpEchoApp.cc:124`**

```cpp
content->removeTagsWherePresent<CreationTimeTag>(b(0), content->getChunkLength());
```

Any other region tag on the echoed content goes with it.

## Why no test catches it

`tests/packet/UnitTest.cc:1793` ("11. removeTagsWherePresent") only ever puts `CreationTimeTag` in the set, so the type-blindness is invisible to it. The reproduction above is that test with a second tag type added.

## Suggested fix

Add a type-aware `clearTags()` overload and call it from `removeTagsWherePresent<T>()`. The four intersection cases are already written; they need the `typeInfo != typeid(*tagObject) -> continue` guard that `getTagIndex()` already uses:

```cpp
void SharingRegionTagSet::clearTags(const std::type_info& typeInfo, b offset, b length);

template<typename T>
inline std::vector<SharingRegionTagSet::RegionTag<T>> SharingRegionTagSet::removeTagsWherePresent(b offset, b length)
{
    auto result = getAllTags<T>(offset, length);
    clearTags(typeid(T), offset, length);
    return result;
}
```

The untyped `clearTags(offset, length)` stays as it is - it is type-blind by design, backing the public `clearTags()` / `clearRegionTags()`.

This is behaviour-changing for anyone relying on the wipe, so it probably wants a WHATSNEW line.

Workaround until then: iterate `getAllTags<T>(offset, length)` and call `removeTag<T>(region.getOffset(), region.getLength())` per region.

MakerYuichi:
Fixes #1184

`removeTagsWherePresent<T>()` filtered its return value by `T` but called the untyped `clearTags(offset, length)`, which wipes every region tag intersecting the range regardless of type. This adds a type-aware `clearTags(typeInfo, offset, length)` overload — reusing the existing intersection logic with the same `typeInfo != typeid(*tagObject) -> continue` guard `getTagIndex()` already uses and calls it from `removeTagsWherePresent<T>()` via `clearTags(typeid(T), ...)`.

The untyped `clearTags(offset, length)` is unchanged, since it's type-blind by design and backs the public `clearTags()`/`clearRegionTags()`.

Added a regression test (`11b` in `testRegionTagSet()`) reproducing the exact scenario from the issue: two different tag types over the same region, removing one type by `T` no longer wipes the other.

### DIFF
--- a/src/inet/common/packet/tag/SharingRegionTagSet.cc
+++ b/src/inet/common/packet/tag/SharingRegionTagSet.cc
@@ -331,5 +331,62 @@ void SharingRegionTagSet::prepareTagsVectorForUpdate()
     }
 }
 
+void SharingRegionTagSet::clearTags(const std::type_info& typeInfo, b offset, b length)
+{
+    if (regionTags != nullptr) {
+        bool changed = false;
+        b clearStartOffset = offset;
+        b clearEndOffset = offset + length;
+        for (size_t i = 0; i < regionTags->size(); i++) {
+            auto& regionTag = (*regionTags)[i];
+            auto tagObject = regionTag.getTag().get();
+            if (typeInfo != typeid(*tagObject))
+                // wrong type, leave it alone
+                continue;
+            else if (clearEndOffset <= regionTag.getStartOffset() || regionTag.getEndOffset() <= clearStartOffset)
+                // no intersection
+                continue;
+            else if (clearStartOffset <= regionTag.getStartOffset() && regionTag.getEndOffset() <= clearEndOffset) {
+                // clear totally covers region
+                prepareTagsVectorForUpdate();
+                regionTags->erase(regionTags->begin() + i--);
+                changed = true;
+            }
+            else if (regionTag.getStartOffset() < clearStartOffset && clearEndOffset < regionTag.getEndOffset()) {
+                // clear splits region into two parts
+                prepareTagsVectorForUpdate();
+                auto& regionTag = (*regionTags)[i];
+                auto startOffset = regionTag.getStartOffset();
+                regionTag.setLength(regionTag.getEndOffset() - clearEndOffset);
+                regionTag.setOffset(clearEndOffset);
+                RegionTag<TagBase> previousRegionTag(regionTag);
+                previousRegionTag.setOffset(startOffset);
+                previousRegionTag.setLength(clearStartOffset - startOffset);
+                regionTags->insert(regionTags->begin() + i++, previousRegionTag);
+                changed = true;
+            }
+            else if (regionTag.getEndOffset() <= clearEndOffset) {
+                // clear cuts end of region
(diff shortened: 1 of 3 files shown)


levy:
Thanks for submitting this! I fixed the raised issue in with a simpler patch in 83396aa6558752d5d3a45458af81a62b08acbd5d

